PR #8, opened 2024-02-12
migration files

leoraba:
- configured tsconfig files on workspace
- Added drizzle and schema files
- generate migration files


prettier.config.js
```diff
@@ -3,7 +3,7 @@ const config = {
 	printWidth: 120,
 	semi: true,
 	singleQuote: true,
-	trailingComma: 'all',
+	trailingComma: 'es5',
```
justincorrigible, line 6: personally, I'm inclined towards "all", curious to hear your opinion for changing this es5

leoraba: no specific reason to use `es5` or `all`. will change to `all` for now.

justincorrigible (approved):

PR #13, opened 2024-02-12
run drizzle migrations

leoraba:
- Added docker-compose
- Added script to run migrations


packages/common/migrations/migrate.ts
```diff
@@ -0,0 +1,13 @@
+import { drizzle } from 'drizzle-orm/postgres-js';
+import { migrate } from 'drizzle-orm/postgres-js/migrator';
+import postgres from 'postgres';
+
+async function main() {
+	const connectionString = process.env.DB_URL || 'postgres://postgres:secret@localhost:5432/lyric';
```
joneubank, line 6: I don't want to be maintaining a hardcoded string into the future, also don't want incorrect credentials being attempted when this is executed in production environments. Can we instead have this throw an error if the env variable `DB_URL` is not available? The error message should contain instructions for what config variables are needed for the migration to run.

joneubank:
```ts
const connectionString = process.env.DB_URL;
if(!connectionString) {
  throw new Error(`No Database connection string is provided. To run migrations there must be an environment variable named 'DB_URL' with the database connection URL (example: postgres://user:password@localhost:5432/dbname).`);
}
```

joneubank (approved):
Looks great, thanks for bringing in the monorepo readme formatting.

We might consider a more concise name for `dictionary_categories`... not sure what the correct name is but this feels quite length for an entity name we are going to reuse frequently. Don't need a change right away, just think about it.

PR #16, opened 2024-02-26
Feat/create new dictionary

leoraba:
Business requirements:
- Store new dictionaries and categories in database using Drizzle ORM
- Fetch dictionaries from Lectern using `lectern-client`

Extras:
- `LyricProvider` to share database connection and other configurations.
- Use `pg.Pool` for database connection pool
- `Common library` sharing express routers and also functions
- Convert project to ESM
- Adding Logger


apps/server/package.json
```diff
@@ -2,21 +2,26 @@
 	"name": "server",
 	"version": "1.0.0",
 	"description": "",
-	"main": "index.js",
+	"main": "dist/index.js",
```
justincorrigible, line 5: 🤔 my first instinct is we shouldn't be pointing at a file in a folder, and have the entry point be at the root level, but then again I'm thinking we could just share the source code instead, and use ts-node.

ultimately there's two different exports or entry points here: as a standalone server, and as dependency library.
meaning we can address this in a later PR.

apps/server/package.json
```diff
@@ -2,21 +2,26 @@
...
 	"scripts": {
 		"build": "rm -rf dist && tsc",
-		"dev": "nodemon server.ts"
+		"dev": "nodemon"
+	},
+	"engines": {
+		"node": ">=18.0.0 <19.0.0"
```
justincorrigible, line 11: current LTS is 20. we shouldn't enforce earlier versions as mandatory.

apps/server/src/config/server.ts
```diff
@@ -0,0 +1,27 @@
+import 'dotenv/config';
+
+import { AppConfig } from 'common';
+
+export const getServerConfig = () => {
+	return {
+		port: process.env.PORT || 3030,
+		debug_mode: process.env.DEBUG || false,
```
justincorrigible, line 8: for boolean env var, this should look like `var.toLowerCase() === 'true'`, so the comparison returns a boolean, and doesn't just assume all strings are turthy e.g. 'false'

apps/server/src/controllers/pingController.ts
```diff
@@ -0,0 +1,7 @@
+import { Request, Response } from 'express';
+
+import pkg from '../../package.json' assert { type: 'json' };
```
justincorrigible, line 3: this is the second import of the manifest, we could abstract it into a reusable manifest.js file that does this import and exports a simplified js object with only the variables we may want to destructure elsewhere

apps/server/src/routes/ping.ts
```diff
@@ -1,5 +1,7 @@
 import { Router } from 'express';
 
+import * as pingController from '../controllers/pingController.js';
```
justincorrigible, line 3: not sure why this `* as var` seems smelly to me. for some reason I feel we should export all the desired items as properties of a default object export in that other file, as if that somehow makes it more explicit for someone looking into what's available there (like a book's "index").

apps/server/src/server.ts
```diff
@@ -2,27 +2,30 @@ import express, { json, urlencoded } from 'express';
...
 const app = express();
 
+app.use(urlencoded({ extended: false }));
+app.use(json());
+
 app.use(helmet());
 
 app.use('/ping', pingRouter);
-app.use('/dictionary', dictionaryRouter);
+app.use('/dictionary', lyricManager.getRouters().getDicionaryRouters());
```
justincorrigible, line 23: do those methods (getRouters, getDictionaryRouters) take parameters? if so, are they documented anywhere?

packages/common/src/config/db.ts
```diff
@@ -0,0 +1,33 @@
+import { drizzle, NodePgDatabase } from 'drizzle-orm/node-postgres';
+import pg from 'pg';
+import { dbInfo } from './config.js';
+
+export class DB {
```
justincorrigible, line 5: we could implement this as a curry function, with an initialiser first layer, that creates and returns an instance to be used where needed. Seems counterintuitive to start using classes here when we're using functional composition everywhere else in the app

packages/common/src/config/logger.ts
```diff
@@ -0,0 +1,30 @@
...
+
+	// file transport
+	if (config.file) {
+		const fileLog = new transports.File({ filename: 'logs.log' });
+		transportList.push(fileLog);
+	}
+
+	const logger = createLogger({
+		level: config.level || 'info',
+		transports: transportList,
```
justincorrigible, line 27: rather than pushing things into an empty array, we could have this line be something that looks more like
```suggestion
		transports: [createConsoleLogger(), createFileLogger()],
```

packages/common/src/controllers/dictionaryController.ts
```diff
@@ -0,0 +1,40 @@
...
+					`[Register Dictionary] Request: categoryName:${categoryName} dictionaryName:${dictionaryName} dictionaryVersion:${dictionaryVersion}`,
+				);
+
+				if (!categoryName) throw new BadRequest('Request is missing `categoryName` parameter.');
+				if (!dictionaryName) throw new BadRequest('Request is missing `dictionaryName` parameter.');
+				if (!dictionaryVersion) throw new BadRequest('Request is missing `version` parameter.');
+
+				const registered = await dictionaryService.registerDictionary(categoryName, dictionaryName, dictionaryVersion);
+				logger.info(`[Register Dictionary] completed!`);
+				res.send(registered);
```
justincorrigible, line 34: a pattern we've been using in other apps to prevent server bugs and make more self-explanatory that the response is the end of an endpoint's functions
```suggestion
				return res.send(registered);
```

packages/common/src/core/manager.ts
```diff
@@ -0,0 +1,27 @@
...
+		return manager;
+	}
+	getRouters() {
+		return {
+			getDicionaryRouters: () => {
+				const routers = dictionaryRouters({
+					db: this.dependencies.db,
+					config: this.dependencies.config,
+					logger: this.dependencies.logger,
+				} as Dependencies);
```
justincorrigible, line 18: rather than asserting "as Type" you can have the function itself return the type, which makes it easier to notice issues within that function if changes are being made whereas the assertion would mask/hide those issues.

packages/common/src/middleware/errorHandler.ts
```diff
@@ -0,0 +1,34 @@
...
+			err.name = 'Not found';
+			customizableMsg = 'Id not found';
+			break;
+		default:
+			status = 500;
+	}
+
+	res.status(status).send({ error: err.name, message: customizableMsg });
+	// pass the error down (so other error handlers can also process the error)
+	next(err);
```
justincorrigible, line 33: This seems like an antipattern. once you've sent a response, there shouldn't be a chance for any other endpoints/middlewares to try to send a response the client won't be listening for

packages/common/src/repository/categoryRepository.ts
```diff
@@ -0,0 +1,41 @@
+import { SelectedFields } from 'drizzle-orm/pg-core/query-builders/select.types';
+import { SQL } from 'drizzle-orm/sql';
+import { isEmpty } from 'lodash-es';
+
+import { GlobalConfig } from '../config/config.js';
+import { Category, NewCategory, dictionaryCategories } from '../models/dictionary_categories.js';
+
+export default class CategoryRepository extends GlobalConfig {
```
justincorrigible, line 8: same comment about classes vs functional composition as before.

looks like what you wanted to do here was something along the lines of

```js
const something = (opts) => {
  const save = async () => {};
  const select = async () => {};

  return {
    save,
    select,
  };
```
...which works great if you want to pass initialisation options to the main function.

otherwise, what you're trying to do is just something like this:

```js
const something = {
  save: async () => {},
  select: async () => {},
};
```

unsure on which one you're aiming for though.

packages/common/src/repository/dictionaryRepository.ts
```diff
@@ -0,0 +1,40 @@
+import { SelectedFields } from 'drizzle-orm/pg-core/query-builders/select.types';
+import { SQL } from 'drizzle-orm/sql';
+import { isEmpty } from 'lodash-es';
+
+import { GlobalConfig } from '../config/config.js';
+import { Dictionary, NewDictionary, dictionaries } from '../models/dictionaries.js';
+
+export default class DictionaryRepository extends GlobalConfig {
```
justincorrigible, line 8: same as the previous comment on classes vs functional composition

packages/common/src/routers/dictionaryRouter.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import { auth } from '../middleware/auth.js';
+
+import { Dependencies } from '../config/config.js';
+import { dictionaryControllers } from '../controllers/dictionaryController.js';
+
+export const dictionaryRouters = (dependencies: Dependencies) => {
+	const router = Router();
+	/**
+	 * @swagger
```
justincorrigible, line 11: 👍

packages/common/src/services/dictionaryService.ts
```diff
@@ -0,0 +1,26 @@
+import { isEmpty } from 'lodash-es';
+import { Dependencies, GlobalConfig } from '../config/config.js';
+
+import getCategoryUtils from '../utils/categoryUtils.js';
+import getDictionaryUtils from '../utils/dictionaryUtils.js';
+
+export default class DictionaryService extends GlobalConfig {
```
justincorrigible, line 7: more classes

packages/common/src/utils/errors.ts
```diff
@@ -0,0 +1,27 @@
+export class BadRequest extends Error {
```
justincorrigible, line 1: never used this pattern, but I can see its usefulness in that other file's "switch"  💡

apps/server/package.json
leoraba: just figured out we won't require this main property in the server/ so I just removed.

apps/server/package.json
leoraba: there's currently and [issue](https://github.com/TypeStrong/ts-node/issues/1997) with ts-node and ESM using node 20. I've tested node versions 18 and 19 and works fine.

justincorrigible: Right! I keep forgetting about that issue… and the more I come across it, the more I feel embracing ESM for our apps is premature.
Never mind, then 🤷

apps/server/src/config/server.ts
leoraba: thanks for catching this. actually removed this debug_mode as there's already a `LOG_LEVEL` we can set as debug when needed

apps/server/src/config/manifest.ts
```diff
@@ -0,0 +1,3 @@
+import pkg from '../../package.json' assert { type: 'json' };
+
+export const { name, version } = pkg;
```
justincorrigible, line 3: 👍 nice encapsulation. all those other files shouldn't have access to the rest of the manifest.
thanks for making this change

apps/server/src/routes/ping.ts
leoraba: fixed this as suggested.

apps/server/src/server.ts
leoraba: changed the `lyricManager` structure, now functions are passed as properties and no need to use `()` empty parenthesis. (i.e. `lyricManager.getRouters.dictionaryRouters`)

packages/common/src/config/logger.ts
leoraba: the reason of pushing into an empty array is that there is condition whether or not adding `File transport log`.  We're checking the condition out of the array

packages/common/src/middleware/errorHandler.ts
leoraba: removed the `next` call as this handler should immediately return the response

packages/common/src/repository/dictionaryRepository.ts
leoraba: changed code to functions

packages/common/src/core/manager.ts
leoraba: changed the manager structure and how dependencies object is passed to functions

apps/server/src/server.ts
```diff
@@ -2,27 +2,30 @@ import express, { json, urlencoded } from 'express';
...
-import dictionaryRouter from './routes/dictionary';
-import pingRouter from './routes/ping';
+import { errorHandler, manager } from 'common';
+import { defaultAppConfig, getServerConfig } from './config/server.js';
+import swaggerDoc from './config/swagger.js';
+import pingRouter from './routes/ping.js';
 
 const serverConfig = getServerConfig();
 
+const lyricManager = manager(defaultAppConfig);
```
justincorrigible, line 12: what does "manager" mean in this context?

leoraba: is a function that creates a manager, it returns an instance to access all the routers, services and other things provided by lyric lib. I guess it should be renamed as `createManager`?

justincorrigible: oh I don't have an issue with it being called this or that (thought `createWhatever` does seem appropriate for that instance creator). I'm just curious about the word "manager" itself, is that meant as a synonym for `router`?

e.g. would this work based on what you're trying to do:`const lyricRouter = createRouter(defaultConfigs);`?

justincorrigible: ah, I see. thanks for explaining: like a Lyric `provider`. that makes sense...  I was going to ask if those "specific functions" rely on the configurations you pass into the "manager" creator, but I can see now how those are all connected.

semantically speaking "manager" is ambiguous and that's why I'm asking about it.

packages/common/src/core/manager.ts
```diff
@@ -0,0 +1,28 @@
...
+
+	return {
+		getConfig: deps,
+		getRouters: {
+			dictionaryRouters: dictionaryRouters(deps),
+		},
+		getFunctions: {
+			dictionaryFunctions: getDictionaryUtils(deps),
+			categoryFunctions: getCategoryUtils(deps),
+		},
```
justincorrigible, line 24: traditionally, `getWhatever` has been used as a function name, so it seems odd naming these properties.
I propose, once instantiated, these properties can be done like `lyricInstance.routers.dictionary()` etc. ("instance" is just a placeholder word instead of "manager" in this case)

```suggestion
		routers: {
			dictionary: dictionaryRouters(deps),
		},
		utils: {
			dictionary: getDictionaryUtils(deps),
			category: getCategoryUtils(deps),
		},
```

apps/server/src/server.ts
```diff
@@ -2,27 +2,30 @@ import express, { json, urlencoded } from 'express';
...
 const app = express();
 
+app.use(urlencoded({ extended: false }));
+app.use(json());
+
 app.use(helmet());
 
 app.use('/ping', pingRouter);
-app.use('/dictionary', dictionaryRouter);
+app.use('/dictionary', lyricManager.getRouters.dictionaryRouters);
```
justincorrigible, line 23: following up on [my suggestion](https://github.com/overture-stack/lyric/pull/16/files#r1509484744)

```suggestion
app.use('/dictionary', lyricProvider.routers.dictionary);
```

packages/common/src/core/manager.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const manager = (configData: AppConfig) => {
+	const deps: Dependencies = {
+		db: connect(configData.db),
+		logger: getLogger(configData.logger),
+		config: configData,
+	};
+
+	return {
+		getConfig: deps,
```
justincorrigible, line 17: hm consecuently with that other suggestion, this would be `.config`

packages/common/src/external/lecternClient.ts
```diff
@@ -0,0 +1,16 @@
+import { entities as dictionaryEntities, restClient as dictionaryRestClient } from '@overturebio-stack/lectern-client';
+
+const client = (schemaServiceUrl: string) => {
+	return {
+		async fetchDictionaryByVersion(name: string, version: string): Promise<dictionaryEntities.SchemasDictionary> {
+			try {
+				const newSchema = await dictionaryRestClient.fetchSchema(schemaServiceUrl, name, version);
+				return newSchema;
+			} catch (error) {
+				throw new Error(`Failed to fetch schema: ${error}`);
```
justincorrigible, line 10: not easy to see from this file, in Github, which other process would catch the error thrown here. would that be the error-handling middleware of an endpoint? or one of the utils?

packages/common/src/repository/categoryRepository.ts
```diff
@@ -0,0 +1,44 @@
...
+		},
+
+		select: async (
+			selectionFields: SelectedFields | undefined,
+			conditions: SQL<unknown> | ((aliases: SelectedFields) => SQL<unknown> | undefined) | undefined,
+		): Promise<
+			| Category[]
+			| {
+					[x: string]: unknown;
+			  }[]
```
justincorrigible, line 29: not a you think nor anything you have to fix, but this styling looks super weird lol

packages/common/src/services/dictionaryService.ts
```diff
@@ -0,0 +1,30 @@
+import { isEmpty } from 'lodash-es';
+import { Dependencies } from '../config/config.js';
+
+import getCategoryUtils from '../utils/categoryUtils.js';
+import getDictionaryUtils from '../utils/dictionaryUtils.js';
+
+const service = (dependencies: Dependencies) => {
+	return {
+		registerDictionary: async (categoryName: string, dictionaryName: string, version: string) => {
```
justincorrigible, line 9: inclined to think when this is imported elsewhere it should go as `*.dictionaryService.register` rather than `*.service.registerDictionary`... unless it is meant to be `*.dictionaryService.registerDictionary`, which sounds redundant, so my question: is this service expected to register anything else but dictionaries?

packages/common/src/external/lecternClient.ts
leoraba: This error is `caught`  by the util/function that is calling this lecternClient, and then thrown to the error handler middleware who is mapping this error with the corresponding HTTP error code.

packages/common/src/services/dictionaryService.ts
leoraba: now that you mention, it makes more sense to name it '*.dictionaryService.register' as there's nothing else to register.

packages/common/src/core/manager.ts
leoraba: will definitely fix naming this properties as suggested. Naming is the most difficult part of coding. thanks!

apps/server/src/config/manifest.ts
joneubank, line 3: Does this work in a single statement?

```suggestion
export { name, version } from '../../package.json' assert { type: 'json' };
```

justincorrigible: good question.
while that would look much simpler, I like the idea that we can do validations or transformations in this file in the future.

packages/common/src/services/dictionaryService.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import getCategoryUtils from '../utils/categoryUtils.js';
+import getDictionaryUtils from '../utils/dictionaryUtils.js';
+
+const dictionaryService = (dependencies: Dependencies) => {
+	return {
+		register: async (categoryName: string, dictionaryName: string, version: string) => {
+			const { createCategoryIfDoesNotExist } = getCategoryUtils(dependencies);
+			const savedCategory = await createCategoryIfDoesNotExist(categoryName);
+			if (isEmpty(savedCategory)) return;
```
joneubank, line 12: Can `createCategoryIfDoesNotExist` even return an empty result? A quick read of the code shows it either throws an error or returns a value...

If it does return an empty value, should we at least log this issue before returning our undefined response?

packages/common/src/services/dictionaryService.ts
```diff
@@ -0,0 +1,30 @@
+import { isEmpty } from 'lodash-es';
+import { Dependencies } from '../config/config.js';
+
+import getCategoryUtils from '../utils/categoryUtils.js';
+import getDictionaryUtils from '../utils/dictionaryUtils.js';
+
+const dictionaryService = (dependencies: Dependencies) => {
+	return {
+		register: async (categoryName: string, dictionaryName: string, version: string) => {
```
joneubank, line 9: So the return type here is `Dictionary | undefined`. Is the dev supposed to interpret `undefined` as a failure to register the dictionary? This appears to be possible because of 2 reasons:
1. unable to get or create the category
2. unable to fetch the dictionary

These are two different reasons for failure that are not being communicated back to the user, and maybe that is acceptable but I am not sure it is.

A bigger issue here is that the two async operations that htis function relies on could both throw an error (as you coded them) but this function does not have a try/catch block to handle this gracefully. This is my largest issue with using thrown errors to communicate issues in TS: there is no check within the code to ensure you catch all thrown errors and handle them completely. Since you are currently throwing errors though we need to be sure to catch them and handle them here. (Make sure to handle each error type from each function separately ;) )

packages/common/src/services/dictionaryService.ts
```diff
@@ -0,0 +1,30 @@
...
+			if (isEmpty(dictionary)) return;
+
+			const savedDictionary = await createDictionaryIfDoesNotExist(dictionaryName, version, savedCategory, dictionary);
+			return savedDictionary;
+		},
+
+		getCurrentDictionary: async () => {
+			// TODO
+			// Read directionary from DB
+			return '';
```
joneubank, line 25: Unimplemented, let's throw an error that makes this clear to make sure any uses of this do not appear to work but actually fail to do anything. Very helpful as the codebase grows.

packages/common/src/utils/categoryUtils.ts
```diff
@@ -0,0 +1,33 @@
...
+const utils = (dependencies: Dependencies) => {
+	return {
+		createCategoryIfDoesNotExist: async (categoryName: string) => {
+			const { logger } = dependencies;
+			try {
+				const categoryRepo = categoryRepository(dependencies);
+				const foundCategory = await categoryRepo.select({}, eq(dictionaryCategories.name, categoryName));
+				if (!isEmpty(foundCategory)) {
+					logger.info(`Category ${categoryName} already exists. Not doing any action`);
+					return foundCategory[0];
```
joneubank, line 17: Unlike the other time I called out array access, we know this is not undefined because of the isEmpty check. This one should be good.

packages/common/src/utils/dictionaryUtils.ts
```diff
@@ -0,0 +1,61 @@
...
+				throw error;
+			}
+		},
+
+		fetchDictionaryByVersion: async (dictionaryName: string, version: string): Promise<SchemasDictionary> => {
+			const { logger } = dependencies;
+			try {
+				const client = lecternClient(dependencies.config.schemaService.url, logger);
+				const dictionaryResponse = await client.fetchDictionaryByVersion(dictionaryName, version);
+				logger.debug(`dictionary fetched from Lectern:${JSON.stringify(dictionaryResponse)}`);
```
joneubank, line 51: add a space between `Lectern:` and the JSON value

joneubank (changes requested):
Awesome changes Leo! The refactoring to an exported provider looks great and is well organized. I found a couple changes that I would like to see changed before we merge but this is great work overall.

A couple general comments to the code I would like to see:
- Exported functions or objects should have JSDocs so that there is in code description of what the function does. This helps importers of our library know what to do with the exported items but also helps future devs understand the scope of each component.
- We need to think through how we're handling failure cases throughout our codebase. Sometimes throwing an error so that the whole operation will fail and the user will see the underlying error is acceptable, but usually we want to communicate the failure with more actionable information so that our code can handle the effect gracefully.
- logger messages should be formatted consistently. If you are going to label the module or function name in some, please make sure those are labelled in all. Keep the text formatting the same wrt. 
  - first letter upper or lower case
  - ending statements in `.` or not
  - how to separate values from the text (put values in quotes, separate with colon with standard amount of spacing, separate with ` - ` or other delimiter)

apps/server/src/config/swagger.ts
```diff
@@ -1,20 +1,20 @@
...
-		version,
+		version: version,
 	},
 };
 
 const options = {
 	swaggerDefinition,
 	// Paths to files containing OpenAPI definitions
-	apis: ['./routes/*.ts'],
+	apis: ['./src/routes/*.ts'],
```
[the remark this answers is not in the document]
justincorrigible: good question! I'd wager the endpoint itself loads at the root level of the router, not in this file. 
though it seems like an odd pattern, it makes sense swagger not to see itself as the center of the app (and therefor the root file for these configs)

apps/server/src/server.ts
```diff
@@ -2,27 +2,30 @@ import express, { json, urlencoded } from 'express';
...
 
 const serverConfig = getServerConfig();
 
+const lyricProvider = provider(defaultAppConfig);
+
 // Create Express server
 const app = express();
 
+app.use(urlencoded({ extended: false }));
+app.use(json());
```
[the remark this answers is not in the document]
justincorrigible: "rather" instead of "also". there shouldn't be any other routers in this app anyway, so that should work just fine

packages/common/src/config/db.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const connect = (info: dbInfo): NodePgDatabase => {
+	const pool = new pg.Pool({
+		host: info.host,
+		port: info.port,
+		database: info.database,
+		user: info.user,
+		password: info.password,
+	});
+	console.log(`Connecting to database on ${info.host}`);
```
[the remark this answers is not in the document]
justincorrigible: winston + morgan, would be my recommendation for next steps on this topic

packages/common/src/controllers/dictionaryController.ts
```diff
@@ -0,0 +1,42 @@
...
+import dictionarySvc from '../services/dictionaryService.js';
+import { BadRequest } from '../utils/errors.js';
+
+const controller = (dependencies: Dependencies) => {
+	const dictionaryService = dictionarySvc(dependencies);
+	return {
+		getCurrentDictionary: async (req: Request, res: Response) => {
+			// const currentDictionary = await dictionaryService.getCurrentDictionary();
+			// res.send(currentDictionary);
+			res.status(200).send();
```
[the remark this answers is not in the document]
justincorrigible: love this one! I'll do that myself from now on too

packages/common/src/controllers/dictionaryController.ts
```diff
@@ -0,0 +1,42 @@
...
+				if (!categoryName) throw new BadRequest('Request is missing `categoryName` parameter.');
+				if (!dictionaryName) throw new BadRequest('Request is missing `dictionaryName` parameter.');
+				if (!dictionaryVersion) throw new BadRequest('Request is missing `version` parameter.');
+
+				const registered = await dictionaryService.register(categoryName, dictionaryName, dictionaryVersion);
+				logger.info(`[Register Dictionary] completed!`);
+				return res.send(registered);
+			} catch (error) {
+				next(error);
+			}
```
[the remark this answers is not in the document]
justincorrigible: inclined to suggest we implement an optional/default handler that they can override with their own.
we wouldn't want to force integrators to have inconsistent handling just for Lyric router vs their own.

packages/common/src/controllers/dictionaryController.ts
```diff
@@ -0,0 +1,42 @@
...
+				const dictionaryVersion = req.body.version;
+
+				logger.info(
+					`[Register Dictionary] Request: categoryName:${categoryName} dictionaryName:${dictionaryName} dictionaryVersion:${dictionaryVersion}`,
+				);
+
+				if (!categoryName) throw new BadRequest('Request is missing `categoryName` parameter.');
+				if (!dictionaryName) throw new BadRequest('Request is missing `dictionaryName` parameter.');
+				if (!dictionaryVersion) throw new BadRequest('Request is missing `version` parameter.');
+
```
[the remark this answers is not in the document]
justincorrigible: would like to see that too. keen on jumping on the Zod bandwagon as well

packages/common/src/repository/categoryRepository.ts
```diff
@@ -0,0 +1,44 @@
...
+		> => {
+			const { db, logger } = dependencies;
+			logger.debug(`Querying Category`);
+			try {
+				if (isEmpty(selectionFields)) return await db.select().from(dictionaryCategories).where(conditions);
```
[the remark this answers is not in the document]
justincorrigible: There's an Eslint rule that reminds the preferred usage (`curly: all` as a warning).
...which reminds me we could add Eslint to this codebase at some point in the near future.

packages/common/src/services/dictionaryService.ts
justincorrigible: just in case, were talking "error 501"

apps/server/src/server.ts
leoraba: renamed to `provider`

packages/common/src/repository/categoryRepository.ts
leoraba: the idea of using `{[x: string]: unknown;}` here is to be able to return whatever the user request on the select command. for example with Drizzle ORM we can request column alias:`db.select({
  field1: users.id,
  field2: users.name,
}).from(category);`  in that case we intend to receive an object with fields {field1: '', field2: ''}. not a category object itself

apps/server/src/config/manifest.ts
leoraba: it looks simpler using a single line, however it doesn't work it throws the error `The requested module '../../package.json' does not provide an export named 'name'`.

apps/server/src/config/swagger.ts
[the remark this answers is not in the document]
leoraba: it's odd indeed. This path is relative to the root directory.

apps/server/src/server.ts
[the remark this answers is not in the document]
leoraba: this is a good point. All lyric provider routers require urlencoded and json parser.

packages/common/src/services/dictionaryService.ts
leoraba: Changed to return only a `Promise<Dictionary>` in case of success. The `undefined` return was removed and now it's handled by a downstream errors. throwing a `ServiceUnavailable` error type in case of failure in downstream services.

apps/server/src/config/server.ts
```diff
@@ -0,0 +1,26 @@
...
+		user: process.env.LYRIC_DB_USER || '',
+		password: process.env.LYRIC_DB_PASSWORD || '',
+	},
+	schemaService: {
+		url: process.env.LECTERN_URL || '',
+	},
+	logger: {
+		level: process.env.LOG_LEVEL || 'info',
+	},
+};
```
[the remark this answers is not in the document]
leoraba: Updated to use a function for required variables, if environment variable is not set throws an error.

leoraba:
Updated PR based on comments.

apps/server/src/config/server.ts
```diff
@@ -9,16 +9,24 @@ export const getServerConfig = () => {
...
+		database: getRequiredConfig('DB_NAME'),
+		user: getRequiredConfig('DB_USER'),
+		password: getRequiredConfig('DB_PASSWORD'),
 	},
 	schemaService: {
-		url: process.env.LECTERN_URL || '',
+		url: getRequiredConfig('LECTERN_URL'),
 	},
 	logger: {
 		level: process.env.LOG_LEVEL || 'info',
```
joneubank, line 32: Good stuff keeping some defaults.

packages/common/src/config/logger.ts
```diff
@@ -1,6 +1,20 @@
...
 import { LoggerConfig } from './config.js';
 
+export type loggerFunction = (...messages: any[]) => void;
+
+export type Logger = {
+	debug: loggerFunction;
+	warn: loggerFunction;
+	info: loggerFunction;
+	error: loggerFunction;
+};
```
joneubank, line 11: Does a default winston logger match this type?

I like this abstraction I just want to make sure winston (which is not the only node logger but is common) can be used as a logger without our own customizations. Don't need further changes for this PR, its just a place we could make further improvements to be more compliant with the JS community.

packages/common/src/controllers/dictionaryController.ts
```diff
@@ -2,35 +2,39 @@ import { NextFunction, Request, Response } from 'express';
...
+	const LOG_MODULE = 'DICTIONARY_CONTROLLER';
 	return {
-		getCurrentDictionary: async (req: Request, res: Response) => {
-			// const currentDictionary = await dictionaryService.getCurrentDictionary();
-			// res.send(currentDictionary);
-			res.status(200).send();
+		getCurrentDictionary: async (req: Request, res: Response, next: NextFunction) => {
+			try {
+				//TODO: implement logic to get current dictionary from DB
+				throw new NotImplemented('This functionallity is not implemented');
```
joneubank, line 15: 👍

packages/common/src/middleware/errorHandler.ts
```diff
@@ -19,6 +33,12 @@ export const errorHandler = (err: Error, req: Request, res: Response, next: Next
 		case err instanceof TSVParseError:
 			status = 422;
 			break;
+		case err instanceof NotImplemented:
+			status = 501;
+			break;
```
joneubank, line 38: 501 👍

joneubank (approved):
Great stuff, looking strong.

packages/common/src/config/logger.ts
leoraba: This is a custom `Logger` type different from winston's type. The purpose of this is to provide 4 standard loggerFunctions, each of these implements a winston function. So, in case we require to use a different type of node logger we will need to implement these 4 logger functions.

PR #19, opened 2024-03-18
Refactor submission

leoraba:
Refactoring:
- Reading only first line of each file to validate required columns before processing whole file
- Parse full TSV file only after all initial batch validations passes
- Put on separate process Schema validation and give a in-process response to user
- Update `State` of submission according schema validation
- Adding `Organization` field on Submissions
- Generate DBML


packages/common/src/utils/fileUtils.ts
```diff
@@ -1,3 +1,4 @@
+import firstline from 'firstline';
```
joneubank, line 1: I see nothing wrong with this module but is it really needed to read one line of a file?

packages/common/src/utils/submissionUtils.ts
```diff
@@ -89,47 +92,53 @@ const utils = (dependencies: Dependencies) => {
...
-						batchName: schemasNames,
-						message: '',
+				if (matchingName.length > 1) {
+					logger.error(LOG_MODULE, `Duplicated schema for file name '${file.originalname}'`);
+					batchErrors.push({
 						type: BATCH_ERROR_TYPE.MULTIPLE_TYPED_FILES,
+						message: 'Multiple schemas matches this file',
+						batchName: file.originalname,
+					});
+				} else if (matchingName.length == 1) {
```
joneubank, line 119: `===`

packages/common/src/utils/types.ts
```diff
@@ -11,24 +14,22 @@ export enum SUBMISSION_STATE {
...
-export type CreateSubmissionResult = {
-	readonly successful: boolean;
-	readonly submission: CreateActiveSubmission;
-	batchErrors: DeepReadonly<BatchError>[];
-};
+export enum CREATE_SUBMISSION_STATE {
+	PROCESSING = 'PROCESSING',
+	INVALID_SUBMISSION = 'INVALID_SUBMISSION',
+	PARTIAL_SUBMISSION = 'PARTIAL_SUBMISSION',
+}
```
joneubank, line 23: I'll provide a code example after I get through the code review... I would like to use TS `enum` as little as possible. As you've written this there are no problems, but it is the only TS code that gets compiled into JS code accessible at run time. This has a few potential weird points. Can leave this for now but remind me to give a JS alternative that provides the same types and values.

joneubank (changes requested):
Really good overall, bunch of small comments and a couple that I'd like to see changed. Hopefully quick fixes and we can review again and approve.

packages/common/docs/schema.dbml
```diff
@@ -0,0 +1,55 @@
...
+
+ref: dictionaries.dictionary_category_id - dictionary_categories.id
+
+ref: submissions.dictionary_category_id - dictionary_categories.id
+
+ref: submissions.dictionary_id - dictionaries.id
+
+ref: submitted_data.dictionary_category_id - dictionary_categories.id
+
+ref: submitted_data.original_schema_id - dictionaries.id
```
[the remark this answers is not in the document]
leoraba: this was solved adding `relationName` property on the model due to having multiple relations between the same two tables  [Ref.](https://orm.drizzle.team/docs/rqb#disambiguating-relations)

packages/common/docs/schema.dbml
```diff
@@ -0,0 +1,55 @@
...
+  dictionary_category_id integer
+  dictionary jsonb
+  created_at timestamp [default: `now()`]
+  created_by varchar
+}
+
+table dictionary_categories {
+  id serial [pk, not null, increment]
+  name varchar [not null, unique]
+  active_dictionary_id integer
```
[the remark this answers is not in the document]
leoraba: added missing relation

packages/common/src/controllers/submissionController.ts
```diff
@@ -16,55 +16,57 @@ const controller = (dependencies: Dependencies) => {
 			try {
 				const categoryId = Number(req.params.categoryId);
 				const files = req.files as Express.Multer.File[];
```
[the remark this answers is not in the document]
leoraba: this is previously handled by the router(submissionRouter.ts). Where we're using `Multer` array function to populate this as an array of Files.

packages/common/src/utils/fileUtils.ts
leoraba: this is to validate if the .tsv contains required columns. so not need to read entire .tsv

apps/server/src/routes/swagger/schemas.yml
```diff
@@ -145,13 +170,14 @@ components:
           type: string
           description: the type of error
           enum:
+            - INVALID_FILE_EXTENSION
             - TSV_PARSING_FAILED
             - INVALID_FILE_NAME
             - INCORRECT_SECTION
             - MULTIPLE_TYPED_FILES
             - UNRECOGNIZED_HEADER
             - MISSING_REQUIRED_HEADER
```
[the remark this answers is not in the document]
leoraba: this was fixed

packages/common/package.json
```diff
@@ -4,8 +4,9 @@
 	"description": "",
 	"main": "dist/index.js",
 	"scripts": {
-		"generate": "pnpm build && drizzle-kit generate:pg",
-		"migrate": "tsx migrations/migrate.ts",
+		"generate": "pnpm build && drizzle-kit generate:pg && pnpm dbmlGen",
+		"migrate": "tsx scripts/migrate.ts",
+		"dbmlGen": "tsx scripts/dbmlGenerator.ts",
```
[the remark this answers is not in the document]
leoraba: updated README file with the description of these scripts

packages/common/src/controllers/submissionController.ts
```diff
@@ -16,55 +16,57 @@ const controller = (dependencies: Dependencies) => {
...
+				const organization = req.body.organization;
```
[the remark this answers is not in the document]
leoraba: added string type to `organization`

packages/common/src/controllers/submissionController.ts
```diff
@@ -16,55 +16,57 @@ const controller = (dependencies: Dependencies) => {
...
 					LOG_MODULE,
-					`Upload Submission Request categoryId '${categoryId}' files '${files?.map((f) => f.originalname)}'`,
+					`Upload Submission Request categoryId '${categoryId}' organization '${organization}' files '${files?.map((f) => f.originalname)}'`,
 				);
 
 				if (isNaN(categoryId)) {
 					throw new BadRequest('Invalid categoryId number format');
 				}
 
+				if (!organization) {
```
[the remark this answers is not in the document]
leoraba: good catch. Added more appropriate validation

packages/common/src/controllers/submissionController.ts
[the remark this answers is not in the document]
leoraba: as per now the path looks like `/submission/category/{categoryId}/upload` and body as form-data requires: files & organization.
I guess adding and organization to the path would be a long URL and not consistent with others endpoint.

packages/common/src/controllers/submissionController.ts
```diff
@@ -16,55 +16,57 @@ const controller = (dependencies: Dependencies) => {
...
-				if (resultSubmission.successful && fileErrors.length == 0 && resultSubmission.batchErrors.length == 0) {
-					status = 201;
+				if (fileErrors.length == 0 && resultSubmission.batchErrors.length == 0) {
 					logger.info(LOG_MODULE, `Submission uploaded successfully`);
 				} else {
 					logger.error(LOG_MODULE, 'Found some errors processing this request');
 				}
+
+				// HTTP 200 status Submission Accepted
+				const status = 200;
```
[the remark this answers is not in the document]
leoraba: this `status` variable is not needed anymore here. I left it here because on previous implementation I was changing it conditionally to return different status. This is now removed to just `res.status(200)`

packages/common/src/models/submissions.ts
```diff
@@ -11,11 +11,14 @@ export const submissions = pgTable('submissions', {
 	id: serial('id').primaryKey(),
 	state: submissionStateEnum('state'),
 	dictionaryCategoryId: integer('dictionary_category_id'),
+	organization: varchar('organization').notNull(),
 	data: jsonb('data').notNull(),
 	errors: jsonb('errors'),
 	dictionaryId: integer('dictionary_id'),
 	createdAt: timestamp('created_at').defaultNow(),
 	createdBy: varchar('created_by'),
+	udpatedAt: timestamp('updated_at').defaultNow(),
```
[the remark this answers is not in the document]
leoraba: Good eye. fixed

packages/common/src/services/submissionService.ts
```diff
@@ -6,95 +6,129 @@ import {
...
+				checkEntityFieldNames,
+				getCurrentActiveSubmission,
+			} = submissionUtils(dependencies);
 			const { getCurrentDictionary } = dictionaryUtils(dependencies);
 
-			let resultSubmission = parseToResultSubmission();
+			let entitiesToProcess: string[] = [];
 			let batchErrors: BatchError[] = [];
 			let submissionSchemaErrors: Record<string, SchemaValidationError[]> = {};
 			let updateSubmissionEntities: Record<string, SubmissionEntity> = {};
```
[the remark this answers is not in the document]
leoraba: changed to use `const` instead of `let`. Thanks for clarifying

packages/common/src/services/submissionService.ts
```diff
@@ -6,95 +6,129 @@ import {
...
 
-			if (submissionsEntities.length > 0) {
+			if (files.length > 0) {
 				const currentDictionary = await getCurrentDictionary(categoryId);
 				const schemasDictionary: SchemasDictionary = {
 					name: currentDictionary.name,
 					version: currentDictionary.version,
 					schemas: currentDictionary.dictionary as SchemaDefinition[],
```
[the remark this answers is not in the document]
leoraba: Fixed. I had forgotten to assign the right type for `dictionary` as in database is stored as a json type the type was unknown. This is fixed now by assigning `.$type<SchemaDefinition[]>()` in the database models. This allows query the database and know the proper type of this field.

packages/common/src/models/submissions.ts
```diff
@@ -11,11 +11,14 @@ export const submissions = pgTable('submissions', {
...
+	updatedBy: varchar('updatedBy'),
```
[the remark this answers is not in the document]
leoraba: sorted all properties but leaving the primary key at the beginning and timestamps at the end to find those easily

packages/common/src/utils/types.ts
leoraba: replaced all typescript enums to use const objects and types instead

apps/server/src/routes/swagger/schemas.yml
```diff
@@ -2,45 +2,70 @@
...
 
   schemas:
     CreateSubmissionResult:
       required:
-        - successful
-        - submission
+        - state
+        - description
         - batchErrors
+        - inProcessEntities
```
justincorrigible, line 49: not asking for changes, but out of curiosity:
does the order of these items change anything (was going to say "why not alphabetical?", but then realised there may be a reason)

packages/common/src/controllers/dictionaryController.ts
```diff
@@ -20,18 +21,18 @@ const controller = (dependencies: Dependencies) => {
...
 					LOG_MODULE,
 					`Register Dictionary Request categoryName '${categoryName}' name '${dictionaryName}' version '${dictionaryVersion}'`,
 				);
 
-				if (!categoryName) throw new BadRequest('Request is missing `categoryName` parameter.');
-				if (!dictionaryName) throw new BadRequest('Request is missing `dictionaryName` parameter.');
-				if (!dictionaryVersion) throw new BadRequest('Request is missing `version` parameter.');
+				if (isEmptyString(categoryName)) throw new BadRequest('Request is missing `categoryName` parameter.');
+				if (isEmptyString(dictionaryName)) throw new BadRequest('Request is missing `dictionaryName` parameter.');
+				if (isEmptyString(dictionaryVersion)) throw new BadRequest('Request is missing `version` parameter.');
```
justincorrigible, line 35: 👍 not not avoiding negative conditions is not bad at all!

packages/common/src/controllers/submissionController.ts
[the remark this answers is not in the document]
justincorrigible: rather than asserting type here, this line could be abstracted to a separate function, wherein we could add validation for the files (?), and then return with the appropriate type

packages/common/src/controllers/submissionController.ts
```diff
@@ -16,57 +17,56 @@ const controller = (dependencies: Dependencies) => {
 			try {
 				const categoryId = Number(req.params.categoryId);
 				const files = req.files as Express.Multer.File[];
+				const organization: string = req.body.organization;
 
 				logger.info(
 					LOG_MODULE,
-					`Upload Submission Request categoryId '${categoryId}' files '${files?.map((f) => f.originalname)}'`,
+					`Upload Submission Request categoryId '${categoryId}' organization '${organization}' files '${files?.map((f) => f.originalname)}'`,
```
justincorrigible, line 24:
```suggestion
					`Upload Submission Request:\n  categoryId '${categoryId}'\n  organization '${organization}'\n  files '${files?.map((f) => f.originalname)}'`,
```

packages/common/src/controllers/dictionaryController.ts
```diff
@@ -20,18 +21,18 @@ const controller = (dependencies: Dependencies) => {
 
 		registerDictionary: async (req: Request, res: Response, next: NextFunction) => {
 			try {
-				const categoryName = req.body.categoryName;
-				const dictionaryName = req.body.dictionaryName;
-				const dictionaryVersion = req.body.version;
+				const categoryName: string = req.body.categoryName;
+				const dictionaryName: string = req.body.dictionaryName;
+				const dictionaryVersion: string = req.body.version;
```
joneubank, line 26: This doesn't do any validation of the content of the `body` object. Since `body` is typed as `any`, `req.body.categoryName` is typed as `any` as well, so TS lets you assign this to `string` even though it could be another object, array, or anything.

When dealing with anything that has type `any` we need to first check the content to make sure it matches our expectations. This can be done using a schema validators (like zod) or through a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#typeof-type-guards)

joneubank (changes requested):

packages/common/src/models/dictionary_categories.ts
```diff
@@ -2,20 +2,28 @@ import { integer, pgTable, serial, timestamp, varchar } from 'drizzle-orm/pg-cor
 
 import { relations } from 'drizzle-orm/relations';
 import { dictionaries } from './dictionaries.js';
+import { submissions } from './submissions.js';
+import { submittedData } from './submitted_data.js';
 
 export const dictionaryCategories = pgTable('dictionary_categories', {
 	id: serial('id').primaryKey(),
+	activeDictionaryId: integer('active_dictionary_id'),
```
justincorrigible, line 10: not sure what the difference is for these two properties.

is the "activeDictionaryId" implying the category has multiple dictionaries that belong to it, but there's going to be a single dictionary in this category that is going to be active?
does that mean other categories will have their own "active" dictionary?

justincorrigible: Discussed over zoom with @leoraba and @joneubank. 
My understanding in the last comment is accurate, and it's just the naming choice of  "dictionary" vs "dictionary category" that may be semantically confusing, and we need to explain through documentation

leoraba: As discussed and just to clarify here, it is right to say that each category is going to have only one active dictionary. Is reasonable to add proper documentation about these terminology.

packages/common/src/controllers/submissionController.ts
```diff
@@ -16,57 +17,56 @@ const controller = (dependencies: Dependencies) => {
...
-							type: BATCH_ERROR_TYPE.TSV_PARSING_FAILED,
 							batchName: file.originalname,
 						};
 
 						fileErrors.push(batchError);
 					}
 				}
 
-				const resultSubmission = await service.uploadSubmission(rawSubmissionEntities, categoryId);
+				const resultSubmission = await service.uploadSubmission(validFiles, categoryId, organization);
```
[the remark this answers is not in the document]
leoraba: function now received an object instead of 3 params

packages/common/src/controllers/submissionController.ts
leoraba: Separating these fields with dashes using logger method

packages/common/src/routers/dictionaryRouter.ts
```diff
@@ -4,13 +4,20 @@ import { auth } from '../middleware/auth.js';
...
 const router = (dependencies: Dependencies): Router => {
 	const router = Router();
 	router.use(urlencoded({ extended: false }));
 	router.use(json());
 
-	router.post('/register', auth, dictionaryControllers(dependencies).registerDictionary);
+	router.post(
+		'/register',
+		auth,
+		validateRequest({ body: registerDictionaryBodyRequestSchema }),
```
justincorrigible, line 18: 👍

packages/common/src/controllers/dictionaryController.ts
joneubank: So I am still nervous about this. For this type to be known both of the following must be true:
- middleware was used on the express path that uses this handler
- middleware must be passed a zod schema with a structure that matches this

Since the request handlers are being written here separately from the express path, there is really no way to ensure this... and you have not put any documentation saying this is necessary and no code checks to prevent it being misused.

As a suggestion for how to fix this, take a look at an example for this approach that we wrote for OHCRN: https://github.com/OHCRN/platform/blob/develop/packages/express-request-validation/src/index.ts

joneubank (changes requested):
I love this middleware approach, but since the type narrowing isn't propagated into the function where the variables are used I still can't be comfortable accepting this. I provided a middleware solution that may be acceptable to you if you want to continue in this path.

packages/common/src/middleware/requestValidation.ts
```diff
@@ -0,0 +1,42 @@
...
+				schema.body.parse(req.body);
+			}
+
+			if (schema.pathParams) {
+				schema.pathParams.parse(req.params);
+			}
+
+			if (schema.files) {
+				schema.files.parse({ files: req.files });
+			}
```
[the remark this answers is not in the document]
leoraba: zod allows custom types and it can be used to check file extension in this case we validate is a .tsv format

leoraba:
Moved the Request validation to a HOC to have the benefit of Typed Request

packages/common/src/utils/requestValidation.ts
```diff
@@ -0,0 +1,51 @@
...
+};
+
+/**
+ * Validate the body, path params or files using Zod parse
+ * @param schema Zod objects used to validate request
+ * @returns Throws a Bad Request when validation fails
+ */
+export function validateRequest<TParams = any, TQuery = any, TBody = any, TFiles = any>(
+	schema: RequestValidation<TParams, TQuery, TBody, TFiles>,
+	handler: RequestHandler<ParamsDictionary, any, TBody>,
```
joneubank, line 20: Since we are validating params and query, we can also specify those types: `RequestHandler<TParams, any, TBody, TQuery>`

joneubank (approved):

packages/common/src/utils/requestValidation.ts
leoraba: Removing Params, Query, files validation as types can not be passed to request

PR #21, opened 2024-03-20
Feat/get active submission

leoraba:
Implementation of GET `/submission/category/{categoryId}` endpoint


Buwujiu:
waiting for approval on refactor PR: https://app.zenhub.com/workspaces/pcgl---pancanadian-genome-library-659eed8bc4fe0c04890cff77/issues/gh/overture-stack/lyric/19

packages/common/src/services/dictionaryService.ts
```diff
@@ -27,7 +29,7 @@ const dictionaryService = (dependencies: Dependencies) => {
 				dictionary.schemas,
 			);
 
-			//TODO: Update the category to have this dictionary as current
+			await categoryRepo.updateCurrentDictionaryOnCategory(savedDictionary.id, savedCategory.id);
```
leoraba, line 32: This update was left in TO DO in previous commits but implemented until now. This is not related to this PR

apps/server/src/routes/swagger/schemas.yml
```diff
@@ -66,62 +66,74 @@ components:
...
-          type: array
-          description: content of the file in JSON format
-        dataErrors:
-          type: array
-          items:
-            $ref: '#/components/schemas/ValidationError'
+          description: The date/time of the last update
+        updatedBy:
+          type: string
+          description: The user who last updated the submission
```
joneubank, line 120: User name or ID? Could be more explicit here.

apps/server/src/routes/swagger/schemas.yml
```diff
@@ -66,62 +66,74 @@ components:
...
+            name:
+              type: string
+            version:
+              type: string
+        dictionaryCategory:
+          type: object
+          description: The Category to which this dictionary belongs
+          properties:
+            id:
+              type: number
```
joneubank, line 134: Do we need the category ID? Just the name should be enough information to work with... unless we have APIs that accept category ID.

apps/server/src/routes/swagger/schemas.yml
```diff
@@ -66,62 +66,74 @@ components:
...
-        createdBy:
+        organization:
           type: string
-          description: the user who created the submission
+          description: Organization the Submission belongs to
+        data:
+          type: object
+          additionalProperties:
+            type: object
+            properties:
```
joneubank, line 87: why is this organized with `additionalProperties`, seems uneccesary?

packages/common/src/controllers/submissionController.ts
```diff
@@ -82,10 +83,20 @@ const controller = (dependencies: Dependencies) => {
 				next(error);
 			}
 		},
-		listActive: async (req: Request, res: Response, next: NextFunction) => {
+		active: async (req: Request, res: Response, next: NextFunction) => {
```
joneubank, line 86: should probably call this `getActive`

packages/common/src/external/lecternClient.ts
```diff
@@ -13,14 +13,16 @@ const client = (schemaServiceUrl: string, logger: Logger) => {
...
-				if (!newSchema) throw new BadRequest(`Schema with name '${name}' and version '${version}' not found`);
-				return newSchema;
+				newSchema = await dictionaryRestClient.fetchSchema(schemaServiceUrl, name, version);
 			} catch (error) {
 				logger.error(LOG_MODULE, `Error Fetching dictionary from lectern`, error);
 				throw new ServiceUnavailable();
 			}
+
+			if (!newSchema) throw new BadRequest(`Schema with name '${name}' and version '${version}' not found`);
+			return newSchema;
```
joneubank, line 25: If you move this inside the try block then we don't need the mutable `newSchema` variable.

packages/common/src/repository/activeSubmissionRepository.ts
```diff
@@ -52,6 +52,50 @@ const repository = (dependencies: Dependencies) => {
 			const updated = await db.update(submissions).set(newData).where(eq(submissions.id, submissionId)).returning();
 			return updated[0];
 		},
+
+		/**
+		 * Get Active Submission by category
+		 * @param {number} categoryId Category ID
+		 * @returns An Active Submission
+		 */
+		getActiveSubmissionWithRelations: async (categoryId: number) => {
```
joneubank, line 61: Can we only have one active submission per category? I would think we could have multiple, one per user...

packages/common/src/services/submissionService.ts
```diff
@@ -142,6 +142,12 @@ const service = (dependencies: Dependencies) => {
 				inProcessEntities: entitiesToProcess,
 			};
 		},
+
+		activeSubmission: async (categoryId: number) => {
```
joneubank, line 146: should probably be called `getActiveSubmission` or similar.

joneubank (changes requested):
Mostly little changes around function naming and swagger descriptions...

I am concerned that we are setting this up to have only a single submission per category, when at very least we want each organization to be able to create a submission per category.

apps/server/src/routes/swagger/schemas.yml
leoraba: Adding `user name` in the description as it is more useful

apps/server/src/routes/swagger/schemas.yml
leoraba: There are APIs that accept category Id. For this reason we're returning the Category ID and also the Category name to identify it more easily.

apps/server/src/routes/swagger/schemas.yml
leoraba: we use `additionalProperties` to refer to property names that are unknown, these properties or "keys" must be string an unique. It represents the `entity name`/`schema name` the records belong to.

packages/common/src/controllers/submissionController.ts
leoraba: Agree. thanks for noticing

packages/common/src/external/lecternClient.ts
leoraba: Left outside of the try block as we want to identify 2 types of error. 
1. The try/catch block throws a ServiceUnavailable when Lectern service is down. 
2. Throw a BadRequest in case dictionary name/version doesn't exist in Lectern.

justincorrigible (commented):
couple changes for consistency

packages/common/src/repository/activeSubmissionRepository.ts
leoraba: this is an important point I missed. We should be able to handle multiple submissions per category at least one per user. Will add userName param.

packages/common/src/services/submissionService.ts
leoraba: renamed to `getActiveSubmission`

leoraba:
Changes applied based on feedback

packages/common/src/external/lecternClient.ts
joneubank: Also since this function is throwing errors I checked for uses of it to make sure they are being properly wrapped in try/catch blocks... there is a usage in register dictionary without a try/catch, this means we could be throwing a `ServiceUnavailable` error from our controller. Do we have a clear response for this type in our error handler?

packages/common/src/controllers/submissionController.ts
```diff
@@ -19,6 +20,10 @@ const controller = (dependencies: Dependencies) => {
 				const categoryId = Number(req.params.categoryId);
 				const files = req.files as Express.Multer.File[];
 				const organization = req.body.organization;
+
+				// TODO: get userName from auth
```
joneubank, line 24: Do we have an issue opened to plan this work?

leoraba: will create a ticket to implement Auth

leoraba:
Adding 3 endpoints to retrieve Active Submissions:
1. `GET /submission/category/{categoryId}` Returns a summary of all Active Submissions by categoryId and user
2. `GET /submission/category/{categoryId}/organization/{organization}` Returns a summary of all Active Submissions by categoryId, organization and user
3. `GET /submission/{submissionId}` Returns an Active Submission

apps/server/src/routes/swagger/submission-api.yml
```diff
@@ -1,8 +1,33 @@
...
+        $ref: '#/components/responses/UnauthorizedError'
+      500:
+        $ref: '#/components/responses/ServerError'
+      503:
+        $ref: '#/components/responses/ServiceUnavailableError'
+
 /submission/category/{categoryId}:
   get:
-    summary: Get active submission for a category
+    summary: Get active submissions for a category
```
joneubank, line 30: Perhaps this is better described as `List active submissions for a category` since it doesn't get the entire active submission.

packages/common/src/repository/activeSubmissionRepository.ts
```diff
@@ -52,6 +95,91 @@ const repository = (dependencies: Dependencies) => {
...
+		 * @param {string} filterParams.userName User Name
+		 * @returns One or many Active Submissions
+		 */
+		getActiveSubmissionsWithRelationsByCategory: async ({
+			categoryId,
+			userName,
+		}: {
+			categoryId: number;
+			userName: string;
+		}): Promise<ActiveSubmissionSummaryRepository[] | undefined> => {
```
joneubank, line 112: If none are found, could this just return an empty array instead of undefined?

joneubank (approved):

apps/server/src/routes/swagger/submission-api.yml
leoraba: Changes as suggested. Thanks.

PR #30, opened 2024-04-15
Query data

leoraba:
Implementation fetching data by category and organization with pagination:
GET `/data/category/{categoryId}` 
GET `/data/category/{categoryId}/organization/{organization}`


packages/data-provider/src/controllers/submittedDataController.ts
```diff
@@ -0,0 +1,69 @@
...
+					throw new BadRequest('Invalid categoryId number format');
+				}
+
+				if (isEmptyString(organization)) {
+					throw new BadRequest('Request is missing `organization` parameter.');
+				}
+
+				const data = await service.getSubmittedDataByOrganization(categoryId, organization, { page, pageSize });
+
+				if (_.isEmpty(data)) throw new NotFound('No Submitted Data found');
```
joneubank, line 59: Should this be more granular? 
- Unknown category
- Unknown organization
- No data for category+organization

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -0,0 +1,41 @@
...
+
+			if (!data || data.length === 0) return;
+
+			return parseSubmittedData(data);
+		},
+		getSubmittedDataByOrganization: async (categoryId: number, organization: string, paginationOps: paginationOps) => {
+			const { getSubmittedDataByOrganizationPaginated } = submittedDataRepo;
+			const { parseSubmittedData } = submittedUtils(dependencies);
+
+			const data = await getSubmittedDataByOrganizationPaginated(categoryId, organization, paginationOps);
```
joneubank, line 27: This function throws an error but that error is not being handled. This is fine when it is invoked from an express router that is protected by an error handler, but if we ever invoke this function from another source this can result in a server crash.

Is this how we want to handle it?

joneubank (changes requested):
I left a couple low priority comments, but I couldn't find the place to put the change I am requesting so I'll summarize here:

Request handlers that perform a paginated query need to include pagination information in the response. This includes the pagination parameters used in the query (page number and page size), but it should also include some information that lets the user know if there are additional pages to search for. Typically this includes the total count for the query.

Can we define a PaginationResponse type that includes the pagination query and response information and that we can union with our other response data to ensure that these new controllers return pagination data in a consistent format?

packages/data-provider/src/controllers/submittedDataController.ts
leoraba: Added validation on the Service layer

packages/data-provider/src/controllers/submittedDataController.ts
```diff
@@ -0,0 +1,69 @@
...
+
+				return res.status(200).send(data);
+			} catch (error) {
+				next(error);
+			}
+		},
+		getSubmittedDataByOrganization: async (req: Request, res: Response, next: NextFunction) => {
+			try {
+				const categoryId = Number(req.params.categoryId);
+				const organization = req.params.organization;
```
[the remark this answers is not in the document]
leoraba: Added parameter definitions in all controllers Request objects. These parameters are treated always as strings and casting to corresponding type is required to be implemented in each controller logic.

packages/data-provider/src/controllers/submittedDataController.ts
```diff
@@ -0,0 +1,69 @@
+import { NextFunction, Request, Response } from 'express';
+import * as _ from 'lodash-es';
+
+import { Dependencies } from '../config/config.js';
+import submittedDataService from '../services/submittedDataService.js';
+import { BadRequest, NotFound } from '../utils/errors.js';
+import { isEmptyString } from '../utils/formatUtils.js';
+
+const controller = (dependencies: Dependencies) => {
```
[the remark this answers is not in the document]
leoraba: Created a type `BaseDependencies` which contains only the database connection and logger instance, this is required for most of the controllers->Services->Repositories.

leoraba:
Changed code as suggested. Ready to merge.

PR #35, opened 2024-04-24
health operation

leoraba:
Implementation of health endpoint. It returns 200 HTTP OK with update server information

### DIFF
--- a/apps/server/src/controllers/healthController.ts
+++ b/apps/server/src/controllers/healthController.ts
@@ -1,0 +1,12 @@
+import { Request, Response } from 'express';
+
+const health = (req: Request, res: Response) => {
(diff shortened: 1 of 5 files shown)


joneubank (commented):
Can we add swagger docs for this?

justincorrigible:
While I don't disagree with the Swagger idea in principle, the benefit of documenting this specific endpoint is not immediately obvious to me.
Care to add some insight to that effect, @joneubank?

joneubank:
Consistency in documenting all our endpoints. In the future someone will want to run this application and need to know what path to use as a health check... they can now reference our API docs.

joneubank (approved):

PR #36, opened 2024-04-29
Feat/separate models

leoraba:
This is a refactoring change to rename/separate `packages/common` folder

```
.
├── apps/
│   └── server
├── packages
│   ├── data-model
│   └── data-provider
```
- `apps/server`: Express Server
- `packages/data-model:` separate drizzle database models into it's own folder, this will facilitate database migration
- `packages/data-provider`:  to include all submission logic. This package is used by server


packages/provider/package.json
```diff
@@ -1,22 +1,24 @@
 {
-	"name": "common",
+	"name": "provider",
```
justincorrigible, line 2: though you were calling it "data-provider"? which is a bit more descriptive then just "provider" (as in, "what does it provide?")
https://github.com/overture-stack/lyric/pull/36/files#diff-c7ffeb6163ab1fc5929df80eda11f4999ae853d5a2bfce072cbf787c02dc0f3cR2

leoraba: it makes more sense `data-provider` will change accordingly

justincorrigible (approved):
PR looks good. thank you for abstracting the `app` from the `common` packages.

Wondering whether we should customise each migration filename to describe what each does (which makes things easier to debug later on if anything goes wrong).

packages/data-model/migrations/meta/_journal.json
```diff
@@ -0,0 +1,13 @@
+{
+  "version": "5",
+  "dialect": "pg",
+  "entries": [
+    {
+      "idx": 0,
+      "version": "5",
+      "when": 1714425272932,
+      "tag": "0000_cute_jimmy_woo",
```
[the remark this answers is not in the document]
leoraba: find out this is now possible to name your migration file on the current beta version [drizzle-kit](https://github.com/drizzle-team/drizzle-kit-mirror/releases/tag/v0.21.0-beta)

PR #40, opened 2024-05-06
Feat/query data filter

leoraba:
Implementation of Query endpoint POST `/data/category/{categoryId}/organization/{organization}/query`
Accepts a `application/json` body with SQON content to query Submitted Data.
Converts SQON to query database.


apps/server/src/server.ts
```diff
@@ -16,6 +16,7 @@ const lyricProvider = provider(defaultAppConfig);
 const app = express();
 
 app.use(helmet());
+app.use(json());
```
leoraba, line 19: Express middleware function to parse json requests

packages/data-provider/src/repository/submittedRepository.ts
```diff
@@ -87,17 +87,24 @@ const repository = (dependencies: BaseDependencies) => {
...
 		 * @param {string} organization Organization Name
+		 * @param {SQL} filter Optional filter
+		 * @param {paginationOps} paginationOps Pagination properties
 		 * @returns The SubmittedData found
 		 */
 		getSubmittedDataByCategoryIdAndOrganizationPaginated: async (
 			categoryId: number,
 			organization: string,
 			paginationOps: paginationOps,
+			filter?: SQL,
```
leoraba, line 98: This function is reusable by adding an optional SQL parameter.

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -51,6 +53,7 @@ const service = (dependencies: BaseDependencies) => {
 			categoryId: number,
 			organization: string,
 			paginationOps: paginationOps,
+			sqon?: SQON,
```
leoraba, line 56: Reusing this function by adding an optional SQON argument

packages/data-provider/src/utils/buildQuery.ts
```diff
@@ -0,0 +1,123 @@
...
+	FilterOperator,
+	GreaterThanFilter,
+	LesserThanFilter,
+	Operator,
+	isArrayFilter,
+	isCombination,
+	isFilter,
+} from '@overture-stack/sqon-builder';
+
+const jsonbColumnName = 'data';
```
leoraba, line 17: Setting this variable with the column name on database of type JSONB. this is used to build queries

apps/server/src/server.ts
justincorrigible: We may want to add an optional limit for those.
In this example I'm giving a fixed 50mb, but this must be something passed as an env var instead.

```suggestion
app.use(json({ limit: '50mb' }));
```

packages/data-provider/src/repository/submittedRepository.ts
justincorrigible: silly question: Are we using SQL in some endpoints and SQON in others? it's unclear to me what makes that difference necessary here.

packages/data-provider/src/utils/buildQuery.ts
justincorrigible: does JSONB stand for json blob? never heard of it until now. may want to add a comment, or rename the variable to be more declarative outside the db context.

packages/data-provider/src/controllers/submittedDataController.ts
```diff
@@ -118,6 +119,77 @@ const controller = (dependencies: BaseDependencies) => {
...
+			req: Request<{ categoryId: string; organization: string }, {}, any, { page: string; pageSize: string }>,
+			res: any,
+			next: any,
+		) => {
+			try {
+				const categoryId = Number(req.params.categoryId);
+				const organization = req.params.organization;
+				const sqon = req.body;
+				let page = parseInt(req.query.page as string) || defaultPage;
+				let pageSize = parseInt(req.query.pageSize as string) || defaultPageSize;
```
justincorrigible, line 133: I don't see these beinf redefined anywhere. Should be const?

packages/data-provider/src/utils/buildQuery.ts
```diff
@@ -0,0 +1,123 @@
...
+	CombinationKeys,
+	CombinationOperator,
+	FilterOperator,
+	GreaterThanFilter,
+	LesserThanFilter,
+	Operator,
+	isArrayFilter,
+	isCombination,
+	isFilter,
+} from '@overture-stack/sqon-builder';
```
justincorrigible, line 15: These should be sorted by import type and separated by an empty line.
Does Lyric include that prettier plugin that helps with this?

packages/data-provider/src/utils/buildQuery.ts
```diff
@@ -0,0 +1,123 @@
...
+
+const iterateOperator = (operator: Operator): SQL<unknown> => {
+	if (isFilter(operator)) {
+		// op in [in, lt, gt]
+		return processFilterOperator(operator);
+	} else if (isCombination(operator)) {
+		// op in [and, or, not]
+		return processCombinationOperator(operator);
+	}
+	throw new BadRequest(`Invalid SQON format. Unsupported SQON Operator`);
```
justincorrigible, line 76: consistent empty spacing 👍 
```suggestion

	throw new BadRequest(`Invalid SQON format. Unsupported SQON Operator`);
```

packages/data-provider/src/utils/buildQuery.ts
```diff
@@ -0,0 +1,123 @@
...
+
+const processCombinationOperator = (sqon: CombinationOperator): SQL<unknown> => {
+	if (sqon.op === CombinationKeys.And) {
+		return andOperator(sqon.content);
+	} else if (sqon.op === CombinationKeys.Or) {
+		return orOperator(sqon.content);
+	} else if (sqon.op === CombinationKeys.Not) {
+		return notOperator(sqon.content);
+	}
+	throw new BadRequest(`Invalid SQON format. Unsupported SQON combination operator: ${sqon}`);
```
justincorrigible, line 91:
```suggestion

	throw new BadRequest(`Invalid SQON format. Unsupported SQON combination operator: ${sqon}`);
```

packages/data-provider/src/utils/buildQuery.ts
```diff
@@ -0,0 +1,123 @@
...
+};
+
+const andOperator = (operations: Operator[]): SQL<unknown> => {
+	const andSql = and(...iterateOperators(operations));
+	if (!andSql) {
+		throw new BadRequest(`Invalid SQON format. Invalid 'and' operator`);
+	}
+	return andSql;
```
justincorrigible, line 99:
```suggestion

	return andSql;
```

packages/data-provider/src/utils/buildQuery.ts
```diff
@@ -0,0 +1,123 @@
...
+};
+
+const orOperator = (operations: Operator[]): SQL<unknown> => {
+	const orSql = or(...iterateOperators(operations));
+	if (!orSql) {
+		throw new BadRequest(`Invalid SQON format. Invalid 'or' operator`);
+	}
+	return orSql;
```
justincorrigible, line 107:
```suggestion

	return orSql;
```

packages/data-provider/src/utils/buildQuery.ts
```diff
@@ -0,0 +1,123 @@
...
+
+const notOperator = (operations: Operator[]): SQL<unknown> => {
+	return not(iterateOperator(operations[0]));
+};
+
+export const convertSqonToQuery = (sqon: Operator | undefined): SQL<unknown> | undefined => {
+	if (!sqon || _.isEmpty(sqon)) {
+		return undefined;
+	}
+	if (isCombination(sqon)) {
```
justincorrigible, line 118: reduce cognitive burden by specifying the two if cases are mutually exclusive
```suggestion
	} else if (isCombination(sqon)) {
```

packages/data-provider/src/utils/types.ts
justincorrigible, line 1: I'll leave TS nonsense to @joneubank 😆

apps/server/src/server.ts
leoraba: `limit` property added on the `SubmittedDataRouter` file. as this is the only router parsing json.

packages/data-provider/src/repository/submittedRepository.ts
leoraba: SQON is a JSON format input by the user in the request. This requires to be transformed to a format understandable by repository layer here is when we use `SQL` that is a type specific for `Drizzle ORM` library.

packages/data-provider/src/utils/dictionaryUtils.ts
```diff
@@ -55,8 +55,8 @@ const utils = (dependencies: Dependencies) => {
 		 */
 		fetchDictionaryByVersion: async (dictionaryName: string, version: string): Promise<SchemasDictionary> => {
 			try {
-				if (!dependencies?.config?.schemaService?.url) throw new Error(`'schemaService' is not configured`);
-				const client = lecternClient(dependencies.config.schemaService.url, logger);
+				if (!dependencies?.schemaService?.url) throw new Error(`'schemaService' is not configured`);
```
justincorrigible, line 58:
```suggestion
				if (!dependencies?.schemaService?.url) {
					throw new Error(`'schemaService' is not configured`);
				}
```

packages/data-provider/src/utils/buildQuery.ts
leoraba: `JSONB` stands for `JSON Binary` ref: [link](https://scalegrid.io/blog/using-jsonb-in-postgresql-how-to-effectively-store-index-json-data-in-postgresql/#:~:text=JSONB%20stands%20for%20%E2%80%9CJSON%20Binary,you%20should%20be%20using%20JSONB.)

Added a comment to this variable to be more declarative

packages/data-provider/src/utils/buildQuery.ts
leoraba: Reorganized imports. somehow prettier didn't organized them earlier...
Lyric has prettier plugin `prettier-plugin-organize-imports` however this doesn't add an empty line. wondering what configuration does this.

packages/data-provider/src/utils/buildQuery.ts
```diff
@@ -0,0 +1,123 @@
...
+			.join(', ');
+	} else if (typeof value === 'string') {
+		// Handle single string
+		return value;
+	} else if (typeof value === 'number') {
+		// Handle single number
+		return value;
+	} else {
+		throw new BadRequest(`Invalid SQON. Unsupported data type: ${typeof value}`);
+	}
```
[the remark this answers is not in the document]
leoraba: sure makes sense

packages/data-provider/src/utils/buildQuery.ts
leoraba: done

packages/data-provider/src/utils/buildQuery.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export const convertSqonToQuery = (sqon: Operator | undefined): SQL<unknown> | undefined => {
+	if (!sqon || _.isEmpty(sqon)) {
+		return undefined;
+	}
+	if (isCombination(sqon)) {
+		// top level SQON must have op in [and, or, not]
+		return processCombinationOperator(sqon);
+	}
+	throw new BadRequest('Invalid SQON format. Top level SQON must be a combination operator (and, or, not');
```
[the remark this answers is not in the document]
leoraba: done

packages/data-provider/src/utils/dictionaryUtils.ts
leoraba: added parenthesis. thanks!

packages/data-provider/src/controllers/submittedDataController.ts
leoraba: changed to `const` as we're not reassigning these variables

leoraba:
Thanks for the suggestions @justincorrigible .

justincorrigible (approved):

apps/server/src/server.ts
joneubank: I think the json middleware has a pretty sane default that I've only ever had to change in order to allow LARGER files.

joneubank (approved):
This is all generally good and to an acceptable standard. Thank you for being so thorough in adding TSDocs.

I did add a couple comments on items that I believe we could handle more cleanly - validating env values and communicating error cases in DB queries.

apps/server/src/server.ts
leoraba: yes, indeed just added limit property to `Multer` middleware, which by default `Multer` has unlimited file size

apps/server/src/config/server.ts
```diff
@@ -28,6 +27,9 @@ export const defaultAppConfig: AppConfig = {
 	schemaService: {
 		url: getRequiredConfig('LECTERN_URL'),
 	},
+	limits: {
+		fileSize: process.env.UPLOAD_LIMIT || '10mb',
+	},
```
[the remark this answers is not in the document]
leoraba: This variable is converted into bytes format using a third party library [bytes](https://www.npmjs.com/package/bytes).  As per it's documentation the only supported units abbreviations are:
- b for bytes
- kb for kilobytes
- mb for megabytes
- gb for gigabytes
- tb for terabytes
- pb for petabytes

Any other text or `typo` is considered as a `byte`. 

As we rely on this library I think we can add this on our README documentation.

packages/data-provider/src/utils/types.ts
```diff
@@ -88,7 +87,10 @@ export type BooleanTrueObject = {
 	[key: string]: true;
 };
 
-export type paginationOps = {
+/**
+ * Pagination Query Params
+ */
+export type PaginationOps = {
```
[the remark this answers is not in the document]
leoraba: Agreed

packages/data-provider/src/repository/submittedRepository.ts
```diff
@@ -62,11 +62,12 @@ const repository = (dependencies: BaseDependencies) => {
...
 		 * Find SubmittedData by category ID with pagination
 		 * @param {number} categoryId Category ID
+		 * @param {PaginationOps} paginationOps Pagination properties
 		 * @returns The SubmittedData found
 		 */
 		getSubmittedDataByCategoryIdPaginated: async (
 			categoryId: number,
-			paginationOps: paginationOps,
+			paginationOps: PaginationOps,
 		): Promise<SubmittedDataRepository[] | undefined> => {
```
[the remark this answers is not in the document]
leoraba: 1. The Controller is appending to the response [in here](https://github.com/overture-stack/lyric/pull/40/files#diff-a57cfe3761e475ef113627066b842e1c07bed84d8c3569183447f7c06a68e528R172) a summary pagination object containing following fields
```
{
	currentPage: number;
	pageSize: number;
	totalPages: number;
	totalRecords: number;
}
```
2. While querying the database it's possible to get `undefined` when there is no records found. However, on Service Layer file there's already a validation in case no records are found on database an error is throw. [see file](https://github.com/overture-stack/lyric/pull/40/files#diff-29f0e5a306a45d631160b0aea0095a665d9b0ad639ab82a8b8624acff3d6f99fR37)

PR #41, opened 2024-05-06
Fix/validate data

leoraba:
- Data validation on Commit should merge submitted Data based on categoryId and organization.
- Data validation on Upload .tsv should merge submitted Data based on categoryId and organization.


packages/data-provider/src/controllers/submittedDataController.ts
```diff
@@ -0,0 +1,120 @@
...
+	return {
+		getSubmittedDataByCategory: async (
+			req: Request<{ categoryId: string }, {}, {}, { pageSize: string; page: string }>,
+			res: Response,
+			next: NextFunction,
+		) => {
+			try {
+				const categoryId = Number(req.params.categoryId);
+				let page = parseInt(req.query.page as string) || defaultPage;
+				let pageSize = parseInt(req.query.pageSize as string) || defaultPageSize;
```
justincorrigible, line 25: considering this variables are not changing, what's preventing from these being defined as constants? (`let` vs `const`)

packages/data-provider/src/controllers/submittedDataController.ts
```diff
@@ -0,0 +1,120 @@
...
+
+				if (_.isNaN(categoryId)) {
+					throw new BadRequest('Invalid categoryId number format');
+				}
+
+				const submittedDataResult = await service.getSubmittedDataByCategory(categoryId, { page, pageSize });
+
+				if (_.isEmpty(submittedDataResult.data)) throw new NotFound('No Submitted Data found');
+
+				return res.status(200).send({
```
justincorrigible, line 49: rather than asserting the type, you could instantiate the response before sending it. e.g.
```js
const response: SubmittedDataPaginatedResponse = {...};

return res.status(200).send(response);
```

packages/data-provider/src/core/provider.ts
```diff
@@ -22,11 +26,14 @@ const provider = (configData: AppConfig) => {
...
-			submission: submissionRouters(deps),
+			submission: submissionRouters(_.omit(deps, ['config'])),
+			submittedData: submittedDataRouters(_.omit(deps, ['config'])),
 		},
 		utils: {
 			dictionary: getDictionaryUtils(deps),
-			category: getCategoryUtils(deps),
+			category: getCategoryUtils(_.omit(deps, ['config'])),
+			submission: getSubmissionUtils(_.omit(deps, ['config'])),
+			submittedData: getSubmittedDataUtils(_.omit(deps, ['config'])),
```
justincorrigible, line 36: looks like a few `omit(deps, ['config'])`... are we expecting to have each of these lines omit different things later on? or could these all be abstracted to a single DRY definition like 
```js
const blah = omit...;

...
routers: { submission: submissionRouters(blah) ... }
```

packages/data-provider/src/repository/submittedRepository.ts
```diff
@@ -1,12 +1,20 @@
...
+const repository = (dependencies: BaseDependencies) => {
 	const LOG_MODULE = 'SUBMITTEDDATA_REPOSITORY';
 	const { db, logger } = dependencies;
+
+	const paginatedColumns = {
+		entityName: true,
+		data: true,
+		organization: true,
+		isValid: true,
+	} as BooleanTrueObject;
```
justincorrigible, line 17: does this not work?

```suggestion
	const paginatedColumns: BooleanTrueObject = {
		entityName: true,
		data: true,
		organization: true,
		isValid: true,
	};
```

packages/data-provider/src/services/submissionService.ts
```diff
@@ -23,60 +22,109 @@ import {
...
+			...activeSubmission?.data,
+			...filesDataProcessed,
+		};
+
+		// This object will merge existing data + new data for validation (Submitted data + active Submission +  tsv parsed Data)
+		const mergeDataRecordsByEntityName = _.mergeWith(
+			mapSubmittedDataSchemaByEntityName(submittedData),
+			mapSubmissionSchemaDataByEntityName(activeSubmission?.id, updatedActiveSubmissionData),
+			(objValue, srcValue) => {
+				if (_.isArray(objValue)) {
```
justincorrigible, line 64:
```suggestion
				if (Array.isArray(objValue)) {
```

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -0,0 +1,110 @@
...
+
+			const isValidCategory = await categoryIdExists(categoryId);
+			if (!isValidCategory) {
+				return {
+					data: [],
+					metadata: {
+						totalRecords: 0,
+						errorMessage: 'Invalid Category ID',
+					},
+				};
```
justincorrigible, line 32: these could be abstracted into a error message dynamic generator

packages/data-provider/src/utils/submissionUtils.ts
```diff
@@ -140,55 +142,70 @@ const utils = (dependencies: Dependencies) => {
...
-		): Promise<{ processedRecords: TypedDataRecord[]; schemaErrors: SchemaValidationError[] }> => {
-			const validRecords: any[] = [];
-			const schemaErrors: any[] = [];
+		determineIfIsSubmission: (
+			toBeDetermined: SubmittedDataReference | SubmissionReference,
+		): toBeDetermined is SubmissionReference => {
+			if ((toBeDetermined as SubmissionReference).type === MERGE_REFERENCE_TYPE.SUBMISSION) {
+				return true;
+			}
+			return false;
```
justincorrigible, line 155: Not a fan of the assertion, but if you must, at least the return can be simplified like so
```suggestion
			return (toBeDetermined as SubmissionReference).type === MERGE_REFERENCE_TYPE.SUBMISSION;
```

packages/data-provider/src/utils/submittedDataUtils.ts
```diff
@@ -4,14 +4,45 @@ import {
...
+					mappingDataRecords[entityName].push({
+						dataRecord: entity.data,
+						reference: {
+							submittedDataId: entity.id,
+							type: MERGE_REFERENCE_TYPE.SUBMITTED_DATA,
+						} as SubmittedDataReference,
+					});
+				});
+			});
+			return mappingDataRecords;
```
justincorrigible, line 44:
```suggestion

			return mappingDataRecords;
```

packages/data-provider/src/controllers/submittedDataController.ts
leoraba: Declaring over asserting 👍

packages/data-provider/src/repository/submittedRepository.ts
leoraba: yes it works!

packages/data-provider/src/utils/submittedDataUtils.ts
leoraba: changed code as suggested!

packages/data-provider/src/services/submissionService.ts
leoraba: changed as suggested

packages/data-provider/src/services/submittedDataService.ts
leoraba: added a function `fetchDataErrorResponse` to return this error response object with the corresponding error message

leoraba:
fix the code as suggested. This is now ready for code review

justincorrigible (approved):
LGTM :shipit: 
Squash it away!

PR #49, opened 2024-06-04
Delete committed submission

### LINKED ISSUE #39: New Endpoint - Delete Committed Submission
## Requirements
- endpoint should delete entity by system ID. 
  - Lyric needs to generate internal system IDs 
- endpoint should have a dry run option default to true, if dry run is enabled, should
return all wil-be deleted entities. 
- if dry run is true, won't delete, need to set dry run to false to execute delete. 
- endpoint should delete entities based on cardinality, there are two relations: 1 to 1(if either one entity is deleted, the other entity will be deleted too); 1 to many (if the "1" is deleted, the 'many' will be deleted; if one of the 'many' is deleted, the '1' remains. ).
- there should be a history or log recording the "reasons' of why these records are deleted. 

## POST MVP Feature
- endpoint should be admin only. (this endpoint will be public until integrated with Auth)

leoraba:
Fix #39 
Implementation of endpoint `DELETE /data/{systemId}`

- Swagger documentation added
- Creates audit records on `audit_submitted_data` database table for every DELETE **Submitted Data records**  keeps track of what was deleted, when and who deleted.
- Query param `dryRun=true` for Dry-Run. Set to `false` to execute deletion. Default is false.
- Finds the dependent relations by using the **foreignKeys restrictrions** structure defined on the Dictionary


apps/server/swagger/data-api.yml
```diff
@@ -148,3 +148,47 @@
...
+        description: If true won't execute the deletion, but will return what will be deleted, default is false.
+        in: query
+        required: false
+        schema:
+          default: false
+          type: boolean
+      - name: reason
+        description: Reason why the record is deleted
+        in: query
+        required: false
```
justincorrigible, line 173: should this be required? or is there a default reason given if one is not provided?

packages/data-model/docs/schema.dbml
```diff
@@ -42,18 +42,21 @@ table submissions {
 
 table submitted_data {
   id serial [pk, not null, increment]
+  comment varchar
```
justincorrigible, line 45: is this the same as "reason"?

packages/data-provider/src/controllers/submittedDataController.ts
```diff
@@ -14,6 +14,44 @@ const controller = (dependencies: BaseDependencies) => {
...
+		deleteSubmittedDataBySystemId: async (
+			req: Request<{ systemId: string }, {}, any, { dryRun: string; reason: string }>,
+			res: any,
+			next: any,
+		) => {
+			try {
+				const systemId = req.params.systemId;
+
+				// Gives false value except when query param is explicitly true
+				const dryRun = req.query.dryRun === 'true' ? true : false;
```
justincorrigible, line 26: because we could support TRUE, or True
```suggestion
				const dryRun = req.query.dryRun?.toLowerCase() === 'true' ? true : false;
```

packages/data-provider/src/repository/submittedRepository.ts
```diff
@@ -1,21 +1,31 @@
...
 import { ServiceUnavailable } from '../utils/errors.js';
-import { BooleanTrueObject, PaginationOptions, SubmittedDataRepository } from '../utils/types.js';
+import { BooleanTrueObject, PaginationOptions, SubmittedDataResponse } from '../utils/types.js';
 
 const repository = (dependencies: BaseDependencies) => {
 	const LOG_MODULE = 'SUBMITTEDDATA_REPOSITORY';
 	const { db, logger } = dependencies;
 
+	// Column name on the database used to build JSONB query
+	const jsonbColumnName = 'data';
```
justincorrigible, line 13: this (as with most constants) could be abstracted into an env var

packages/data-provider/src/repository/submittedRepository.ts
```diff
@@ -1,21 +1,31 @@
...
 		data: true,
 		organization: true,
 		isValid: true,
 		systemId: true,
 	};
+
+	// Enable or disable SoftDelete filter
+	// Manually adding softDelete filter as Drizzle currently doesn't support Soft-Delete option
+	// Important: Make sure all Queries contains this filter
+	const softDelete = true;
```
justincorrigible, line 26: same for this (env var)

packages/data-provider/src/repository/submittedRepository.ts
```diff
@@ -98,13 +112,14 @@ const repository = (dependencies: BaseDependencies) => {
 			organization: string,
 			paginationOptions: PaginationOptions,
 			filter?: SQL,
-		): Promise<SubmittedDataRepository[] | undefined> => {
+		): Promise<SubmittedDataResponse[] | undefined> => {
```
justincorrigible, line 115: is the undefined return necessary here?

packages/data-provider/src/services/submissionService.ts
```diff
@@ -321,6 +321,7 @@ const service = (dependencies: BaseDependencies) => {
 						data: record,
 						dictionaryCategoryId: categoryId,
 						entityName,
+						isValid: false, // default as false
```
justincorrigible, line 324: this could use a comment explaining where it gets redefined as true

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -73,7 +133,7 @@ const service = (dependencies: BaseDependencies) => {
 				paginationOptions,
 				filterSql,
 			);
-			if (!recordsPaginated) {
+			if (!notEmpty(recordsPaginated)) {
```
justincorrigible, line 136: is this `not notEmpty` meant to somehow be better than `lodash.isEmpty`?

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -85,12 +145,60 @@ const service = (dependencies: BaseDependencies) => {
...
 
 			return {
-				data: parseSubmittedData(recordsPaginated),
+				data: recordsPaginated,
 				metadata: {
 					totalRecords,
 				},
 			};
 		},
+		deleteSubmittedDataBySystemId: async (
```
justincorrigible, line 154: for consistency with the other files
```suggestion

		deleteSubmittedDataBySystemId: async (
```

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -58,7 +118,7 @@ const service = (dependencies: BaseDependencies) => {
 			const { getSubmittedDataByCategoryIdAndOrganizationPaginated, getTotalRecordsByCategoryIdAndOrganization } =
 				submittedDataRepo;
 			const { categoryIdExists } = categoryRepository(dependencies);
-			const { parseSubmittedData, fetchDataErrorResponse } = submittedUtils(dependencies);
+			const { fetchDataErrorResponse } = submittedUtils(dependencies);
 
 			const isValidCategory = await categoryIdExists(categoryId);
 			if (!isValidCategory) {
```
justincorrigible, line 124:
```suggestion

			if (!isValidCategory) {
```

packages/data-provider/src/services/submittedDataService.ts
justincorrigible:
```suggestion

			if (!notEmpty(recordsPaginated)) {
```

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -85,12 +145,60 @@ const service = (dependencies: BaseDependencies) => {
...
+			reason: string,
+			userName: string,
+		): Promise<SubmittedDataResponse[]> => {
+			const { getSubmittedDataBySystemId } = submittedDataRepo;
+			const { getDictionaryById } = dictionaryRepo;
+			const { mapRecordsSubmittedDataResponse } = submittedUtils(dependencies);
+
+			// get SubmittedData by SystemId
+			const submittedData = await getSubmittedDataBySystemId(systemId);
+			if (!notEmpty(submittedData)) {
```
justincorrigible, line 166:
```suggestion

			if (!notEmpty(submittedData)) {
```

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -85,12 +145,60 @@ const service = (dependencies: BaseDependencies) => {
...
+				throw new BadRequest(`No Submitted data found with systemId '${systemId}'`);
+			}
+			logger.info(LOG_MODULE, `Found Submitted Data with system ID '${systemId}'`);
```
justincorrigible, line 169:
```suggestion

			logger.info(LOG_MODULE, `Found Submitted Data with system ID '${systemId}'`);
```

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -85,12 +145,60 @@ const service = (dependencies: BaseDependencies) => {
...
+
+			// create array with records to be updated
+			const recordsToUpdate: SubmittedData[] = [submittedData];
+
+			// get dictionary
+			const dictionary = await getDictionaryById(submittedData.lastValidSchemaId);
+			if (!dictionary) {
```
justincorrigible, line 176:
```suggestion

			if (!dictionary) {
```

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -85,12 +145,60 @@ const service = (dependencies: BaseDependencies) => {
...
+				throw new BadRequest(`Dictionary not found`);
+			}
+			// get dictionary relations
```
justincorrigible, line 179:
```suggestion

			// get dictionary relations
```

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -85,12 +145,60 @@ const service = (dependencies: BaseDependencies) => {
...
+			const dictionaryRelations = getDictionarySchemaRelations(dictionary);
+
+			const recordDependency = await searchDirectDependents(dictionaryRelations, submittedData);
+			if (recordDependency && recordDependency.length > 0) {
```
justincorrigible, line 183:
```suggestion

			if (recordDependency && recordDependency.length > 0) {
```

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -85,12 +145,60 @@ const service = (dependencies: BaseDependencies) => {
...
+			if (dryRun === false) {
+				// Execute soft deletion on Submitted Data
+				const submittedDataIds = recordsToUpdate.map((records) => records.id);
+				const updatedRecords = await submittedDataRepo.updateMany(submittedDataIds, {
+					comment: `Soft-Delete reason: ${reason}`,
+					deletedAt: new Date(),
+					deletedBy: userName,
+				});
+				logger.info(LOG_MODULE, `Successfully soft deleted Submitted Data. Total records '${updatedRecords.length}'`);
+				return mapRecordsSubmittedDataResponse(updatedRecords);
```
justincorrigible, line 196:
```suggestion

				logger.info(LOG_MODULE, `Successfully soft deleted Submitted Data. Total records '${updatedRecords.length}'`);

				return mapRecordsSubmittedDataResponse(updatedRecords);
```

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -85,12 +145,60 @@ const service = (dependencies: BaseDependencies) => {
...
+
+			if (dryRun === false) {
+				// Execute soft deletion on Submitted Data
+				const submittedDataIds = recordsToUpdate.map((records) => records.id);
+				const updatedRecords = await submittedDataRepo.updateMany(submittedDataIds, {
+					comment: `Soft-Delete reason: ${reason}`,
+					deletedAt: new Date(),
+					deletedBy: userName,
+				});
+				logger.info(LOG_MODULE, `Successfully soft deleted Submitted Data. Total records '${updatedRecords.length}'`);
```
justincorrigible, line 195:
```suggestion

				logger.info(LOG_MODULE, `Successfully soft deleted Submitted Data. Total records '${updatedRecords.length}'`);
```

packages/data-provider/src/utils/dictionarySchemaRelations.ts
```diff
@@ -0,0 +1,43 @@
...
+ * @param {Dictionary} dictionary
+ * @returns {Record<string, SchemaNode[]>}
+ */
+export const getDictionarySchemaRelations = (dictionary: Dictionary): Record<string, SchemaNode[]> => {
+	const dictionaryRelations: Record<string, SchemaNode[]> = {};
+
+	dictionary.dictionary.map((schemaDefinition: SchemaDefinition) => {
+		const childrenSchemaName = schemaDefinition.name;
+
+		schemaDefinition.restrictions?.foreignKey?.map((foreignKey) => {
```
justincorrigible, line 23:
```suggestion
		schemaDefinition.restrictions?.foreignKey?.map((foreignKey) => {
```

packages/data-provider/src/utils/submittedDataUtils.ts
```diff
@@ -101,6 +95,23 @@ const utils = (dependencies: BaseDependencies) => {
...
+		mapRecordsSubmittedDataResponse: (submittedData: SubmittedData[]): SubmittedDataResponse[] => {
+			return submittedData.map((data) => {
+				return {
+					data: data.data,
+					entityName: data.entityName,
+					isValid: data.isValid,
+					organization: data.organization,
+					systemId: data.systemId,
+				};
+			});
```
justincorrigible, line 112: Unless there's other manipulation meant to happen before the return, you could use the implicit one.
```suggestion
			return submittedData.map((data) => ({
				data: data.data,
				entityName: data.entityName,
				isValid: data.isValid,
				organization: data.organization,
				systemId: data.systemId,
			}));
```

packages/data-model/docs/schema.dbml
leoraba: Yes. `comment` column stores the "reason" for a deletion

apps/server/swagger/data-api.yml
leoraba: There is no default `reason`. In a happy scenario the user should be the responsible to define the reason for a deletion, however will leave this query param as **required** as a safe check

packages/data-provider/src/repository/submittedRepository.ts
leoraba: Removed `undefined` as it's not necessary. In case there's no records found querying the database using Drizzle function `findMany` it will return `[]`

packages/data-provider/src/services/submissionService.ts
leoraba: added comment in the code
`By default New Submitted Data is created as invalid until validation process proves otherwise`

packages/data-provider/src/repository/submittedRepository.ts
leoraba: removed this variable as soft delete is required, there is no other option other than soft delete.

leoraba:
Finished code refactor as suggested on comments

packages/data-provider/src/repository/submittedRepository.ts
```diff
@@ -1,21 +1,29 @@
...
 		entityName: true,
 		data: true,
 		organization: true,
 		isValid: true,
 		systemId: true,
 	};
+
+	// Adding softDelete filter as Drizzle currently doesn't support Soft-Delete option
+	// Important: Make sure all Queries on this repository contains this filter
+	const softDeleteFilter = isNull(submittedData.deletedAt);
```
joneubank, line 25: Is there any way to enforce that this is always used? This comment does not protect future devs from making a mistake with this.

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -73,24 +136,82 @@ const service = (dependencies: BaseDependencies) => {
...
+
+			if (recordDependency && recordDependency.length > 0) {
+				recordsToUpdate.push(...recordDependency);
+			}
+
+			if (dryRun === false) {
+				// Execute soft deletion on Submitted Data
+				const submittedDataIds = recordsToUpdate.map((records) => records.id);
+				const updatedRecords = await submittedDataRepo.updateMany(submittedDataIds, {
+					comment: `Soft-Delete reason: ${reason}`,
```
joneubank, line 201: Will this property ever contain information other than this message? Cause if not we should just call it `deleteReason`.

I'm wondering if we should revisit adding a "RecordChangeHistory" table where we can record "update" and "delete" actions.

packages/data-provider/src/utils/dictionarySchemaRelations.ts
```diff
@@ -0,0 +1,41 @@
...
+						schemaName: parentSchemaName,
+						fieldName: mapping.foreign,
+					},
+				};
+
+				dictionaryRelations[parentSchemaName] = dictionaryRelations[parentSchemaName] || [];
+				dictionaryRelations[parentSchemaName].push(childrenNode);
+			});
+		});
+	});
```
joneubank, line 38: This `.map` is not returning anything, but looks to want to be building the contents of the object `dictionaryRelations`... This is a `reduce` operation.

If you want to leave it written this way, change the `.map` to a `.forEach`. Since you aren't actually returning anything there's no need to build the array that `.map` returns.

leoraba: refactoring this function as suggested

packages/data-provider/src/services/submittedDataService.ts
leoraba: renaming this column from `comment` to `deleteReason`

apps/server/swagger/data-api.yml
```diff
@@ -148,3 +148,47 @@
...
+          type: string
+      - name: dryRun
+        description: If true won't execute the deletion, but will return what will be deleted, default is false.
+        in: query
+        required: false
+        schema:
+          default: false
+          type: boolean
+      - name: reason
+        description: Reason why the record is deleted
```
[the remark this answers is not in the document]
leoraba: This is a required field. The only validation is string should not be empty.

packages/data-model/migrations/meta/0003_snapshot.json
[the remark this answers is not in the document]
leoraba: The snapshot files should be included in git as they are used by Drizzle to build new migration (.sql files) incrementally.

packages/data-provider/src/repository/submittedRepository.ts
leoraba: Currently, Drizzle doesn't support soft-delete natively,  so far I haven't found an alternative way to handle soft-delete queries, other than manually adding this `softDeleteFilter` filter.

packages/data-provider/src/services/submittedDataService.ts
leoraba: An approach could be to have a table like "RecordChangeHistory" along with a trigger so AFTER an update INSERT into  RecordChangeHistory table. All this should be implemented on a separate PR for separation of concerns

packages/data-model/docs/schema.dbml
leoraba: changed to column name `deleteReason`

packages/data-provider/src/repository/submittedRepository.ts
leoraba: this value is now imported from the `data-model`

leoraba:
Code fixed as suggested on comments. This is now ready for code review.

packages/data-provider/src/repository/submittedRepository.ts
joneubank: I mean more like, as a developer every query I write needs to have this filter. How do we ensure that no code is committed that has a query without this filter? Is there some code pattern or abstraction that would help?

What you have here works until some new dev tries to edit this file in 3 years and writes a new query without reading this comment.

joneubank (approved):
Great addition for soft delete.

The schema relations graph construction may be useful to pull into the Lectern schema. Once that code stabilizes we should discuss it.

packages/data-provider/src/utils/dictionarySchemaRelations.ts
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Returns all the children and it's relations by each schema on a Dictionary
+ * @param {Dictionary} dictionary
+ * @returns {Record<string, SchemaChildNode[]>}
+ */
+export const getDictionarySchemaRelations = (dictionary: Dictionary): Record<string, SchemaChildNode[]> => {
+	const dictionaryRelations = dictionary.dictionary.reduce(
+		(acc: Record<string, SchemaChildNode[]>, schemaDefinition: SchemaDefinition) => {
+			schemaDefinition.restrictions?.foreignKey?.forEach((foreignKey) => {
+				const parentSchemaName = foreignKey.schema;
```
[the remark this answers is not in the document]
leoraba: Just refactor this function to use nested `reduce` functions. The mapping accumulator is created  in the inner function and passed to the outer.

apps/server/swagger/data-api.yml
leoraba: renamed to `comment` this value is stored as a `comment` on an Audit table

packages/data-model/migrations/0003_set_not_null_columns_submitted_data.sql
leoraba, line 1: Not Delete Submitted Data related, so creating a separate migration .sql file

packages/data-model/migrations/meta/0003_snapshot.json
leoraba, line 1: Snapshots file are used by Drizzle to create migration incrementally

packages/data-provider/src/repository/submittedRepository.ts
```diff
@@ -16,7 +43,18 @@ const repository = (dependencies: BaseDependencies) => {
 		isValid: true,
 		systemId: true,
 	};
+
 	return {
+		delete: async (data: SubmittedData, comment: string, userName: string) => {
+			const deletedRecord = await db.delete(submittedData).where(eq(submittedData.id, data.id));
+
+			if (audit.enabled) {
```
leoraba, line 51: Reading from a env variable. default is true, but if set to false no audit records are stored

leoraba (commented):
Did a refactor on Delete Submitted Data.
This Endpoint will delete permanently the Submitted Data by SystemID and all data depending on it. And create a record on a separate table `audit_submitted_data` this will keep all the log information of the previous data, who, why and when this happened.

README.md
```diff
@@ -74,18 +74,19 @@ The Environment Variables used for this application are listed in the table bell
...
 | `LOG_LEVEL`          | Log Level                                                                                                                                                                                                                                                                                            | 'info'                                 |
+| `PORT`               | Server Port.                                                                                                                                                                                                                                                                                         | 3030                                   |
+| `UPLOAD_LIMIT`       | Limit upload file size in string or number. <br>Supported units and abbreviations are as follows and are case-insensitive: <br> - b for bytes<br> - kb for kilobytes<br>- mb for megabytes<br>- gb for gigabytes<br>- tb for terabytes<br>- pb for petabytes<br>Any other text is considered as byte | '10mb'                                 |
```
joneubank, line 89: woah, html formatting. so fancy.

packages/data-provider/src/config/config.ts
```diff
@@ -39,12 +43,14 @@ export type AppConfig = {
 	logger: LoggerConfig;
 	schemaService: SchemaServiceConfig;
 	idService: IdServiceConfig;
+	audit: AuditConfig;
```
joneubank, line 46: Since this AppConfig is part of the public interface for the Lyric Data-Pprovider, I am wondering if there is a better way to organize this feature. My two thoughts are:
1. Since we have default behaviour (audit is enabled) could we make this an optional part of the config, where people would need to intentionally disable it?
2. We can expect to have more optional features in the future, so instead of having a whole property just for `audit`, should we instead group these inside a property called `features`? Something like:

```ts
export type AppConfig = {
	db: DbConfig;		db: DbConfig;
	limits: LimitsConfig;		limits: LimitsConfig;
	logger: LoggerConfig;		logger: LoggerConfig;
	schemaService: SchemaServiceConfig;		schemaService: SchemaServiceConfig;
	idService: IdServiceConfig;		idService: IdServiceConfig;
	features?: { audit?: AuditConfig; }
}
```

joneubank: Also, please sort the properties in this type alphabetically.

packages/data-provider/src/config/config.ts
```diff
@@ -39,12 +43,14 @@ export type AppConfig = {
...
 };
 
 /**
  * Base Dependencies required for utils/services
  */
 export interface BaseDependencies {
+	audit: AuditConfig;
```
joneubank, line 53: This is in the correct place alphabetically but the location of `idService` is undoing your hard work ;P

joneubank (changes requested):
Left several suggestions. I'll trust you to decide which of these are good ideas and which are not necessary.

apps/server/src/config/server.ts
```diff
@@ -25,7 +28,7 @@ export const defaultAppConfig: AppConfig = {
 		password: getRequiredConfig('DB_PASSWORD'),
 	},
 	idService: {
-		useLocal: process.env.ID_USELOCAL === 'true',
+		useLocal: (process.env.ID_USELOCAL ?? 'true').toLocaleLowerCase() === 'true',
```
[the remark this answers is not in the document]
leoraba: created a `getBoolean` function to include this logic

apps/server/src/config/server.ts
```diff
@@ -17,6 +17,9 @@ const getRequiredConfig = (name: string) => {
 };
 
 export const defaultAppConfig: AppConfig = {
+	audit: {
+		enabled: (process.env.AUDIT_ENABLED ?? 'true').toLocaleLowerCase() === 'true',
```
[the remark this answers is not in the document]
leoraba: function created.

packages/data-provider/src/config/config.ts
leoraba: sorted alphabetically.

packages/data-provider/src/config/config.ts
leoraba: I've reorganized audit properties on to features, this will allow add additional features in the future

packages/data-model/docs/schema.dbml
```diff
@@ -6,6 +11,23 @@ enum submission_status {
...
 }
 
+table audit_submitted_data {
+  id serial [pk, not null, increment]
+  action audit_action [not null]
+  comment varchar
+  dictionary_category_id integer [not null]
+  entity_name varchar [not null]
+  is_valid boolean [not null]
+  last_valid_schema_id integer [not null]
```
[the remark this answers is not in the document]
leoraba: Updated this table to keep track of the validity of old and new data.
```
new_data_is_valid boolean [not null]
old_data_is_valid boolean [not null]
```

leoraba:
Updated PR addressing all the comments. This PR is ready for code review

packages/data-model/migrations/0004_audit_submitted_data.sql
```diff
@@ -10,10 +10,11 @@ CREATE TABLE IF NOT EXISTS "audit_submitted_data" (
 	"comment" varchar,
 	"dictionary_category_id" integer NOT NULL,
 	"entity_name" varchar NOT NULL,
-	"is_valid" boolean NOT NULL,
 	"last_valid_schema_id" integer NOT NULL,
 	"new_data" jsonb,
+	"new_data_is_valid" boolean NOT NULL,
```
joneubank, line 15: Since this is `NOT NULL`, What is the value here when `new_data` is null?

joneubank (approved):

packages/data-model/migrations/0004_audit_submitted_data.sql
leoraba: in the case where `new_data` is null `new_data_is_valid` will be null. This is the case of Delete submitted data

Issue #62, opened 2024-06-20
Get Categories Endpoint


Buwujiu:
An endpoint to list all categories, should return:
- category names
- organization names under each category
- dictionary under each category + dic name + dict version

Buwujiu:
Leo to add organization names under each category to existing PR.

Buwujiu:
Tested in Dev, working as expected.

PR #65, opened 2024-06-24
List categories endpoint

leoraba:
Get categories Endpoint issue #62


apps/server/src/server.ts
```diff
@@ -17,10 +17,14 @@ const app = express();
 
 app.use(helmet());
 
+// Ping Route
 app.use('/ping', pingRouter);
+
+// Lyric Routes
+app.use('/category', lyricProvider.routers.category);
```
leoraba, line 24: new Endpoint to list categories

packages/data-provider/src/core/provider.ts
leoraba, line 1: exporting routers, services and utils. Allows other devs to resuse existing Lyric functions

packages/data-provider/src/services/categoryService.ts
```diff
@@ -0,0 +1,43 @@
...
+					id: category.id,
+					dictionary: {
+						name: category.activeDictionary?.name || '',
+						version: category.activeDictionary?.version || '',
+					},
+					name: category.name,
+					organizations: organizationsFound,
+					createdAt: _.toString(category.createdAt?.toISOString()),
+					createdBy: _.toString(category.createdBy),
+					updatedAt: _.toString(category.updatedAt),
```
joneubank, line 29: Does this need to be toISOString like the createdAt property?

packages/data-provider/src/services/categoryService.ts
```diff
@@ -0,0 +1,43 @@
...
+					updatedBy: _.toString(category.updatedBy),
+				};
+			}
```
joneubank, line 32: What is the reason for converting so many of these values to different types form how they are actually stored? I do like that we have a different `Category` type from `CategoryDetailsResponse`, and I understand converting dates to strings,  but it seems weird that we are converting some values from `undefined` to `''`. This is happening within the `dictionary` property, so I assume this is a quirk of the situation where a category could fail to join with an `activeDictionary`? When `category.activeDictionary` is undefined, perhaps just return `dictionary: undefined` instead?

packages/data-provider/src/services/categoryService.ts
leoraba: Missed to add `.toISOString()`. Will fix this.

packages/data-provider/src/repository/submittedRepository.ts
```diff
@@ -80,6 +80,26 @@ const repository = (dependencies: BaseDependencies) => {
...
 
+		/**
+		 * Returns a list of all organizations found by category ID
+		 * @param {number} categoryId
+		 * @returns
+		 */
+		getAllOrganizationsByCategoryId: async (categoryId: number): Promise<string[]> => {
+			try {
+				const resultQuery = await db
+					.selectDistinct({ organization: submittedData.organization })
```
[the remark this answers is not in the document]
leoraba: Adding an index by organization will definitely improve performance on this query.

packages/data-provider/src/services/categoryService.ts
leoraba: The only reason why this conversion of `name` and `version` to `''` is when a join between the category and activeDictionary fails, as you mentioned in your comment and it would make more sense to change dictionary as optional as suggested.

leoraba:
PR updated to solve previous comments.

packages/data-provider/src/services/categoryService.ts
```diff
@@ -18,15 +18,17 @@ const categoryService = (dependencies: BaseDependencies) => {
...
-					dictionary: {
-						name: category.activeDictionary?.name || '',
-						version: category.activeDictionary?.version || '',
-					},
+					...(category.activeDictionary && {
+						dictionary: {
+							name: category.activeDictionary.name,
+							version: category.activeDictionary.version,
+						},
+					}),
```
joneubank, line 26: This is correct but was hard for me to read, would you consider instead:
```suggestion
					dictionary: category.activeDictionary
						? {
								name: category.activeDictionary.name,
								version: category.activeDictionary.version,
							}
						: undefined,
```

joneubank (approved):

packages/data-provider/src/services/categoryService.ts
leoraba: formatting code as suggested.

Issue #60, opened 2024-06-20
BUG :bug: - Register Existing Dictionary on a new category doesn't return new category 


leoraba:
## Description
Issue found on `POST /dictionary/register` endpoint.
When Registering an existing Dictionary(name & version) on a new Category it doesn't return the new category ID

## Expected Behaviour
Response should include the new CategoryID

## Actual Behaviour
Response returns old CategoryID

Buwujiu:
Tested in Dev, working as expected!

Buwujiu:
Tested in Dev, working as expected.

PR #68, opened 2024-07-04
Fix/issue 60 return new category

leoraba:
**This PR contains:**
- Fix issue #60 
- Swagger docs improvements: Add missing schemas and reorganize schemas alphabetically


apps/server/swagger/schemas.yml
leoraba, line 1: Just sorting in alphabetical order

apps/server/swagger/schemas.yml
```diff
@@ -317,6 +344,25 @@ components:
...
+        dictionary:
+          type: object
+          description: Schema of the Dictionary
+        name:
+          type: string
+          description: Name of the Dictionary
+        version:
+          type: string
+          description: Version of the Dictionary
+
```
leoraba, line 365: Adding this Schema as it was missing

packages/data-model/docs/schema.dbml
```diff
@@ -32,7 +32,6 @@ table audit_submitted_data {
 table dictionaries {
   id serial [pk, not null, increment]
   dictionary jsonb [not null]
-  dictionary_category_id integer
```
leoraba, line 35: Removing this reference of a category.
A dictionary can have many categories

packages/data-provider/src/controllers/dictionaryController.ts
```diff
@@ -31,9 +32,22 @@ const controller = (dependencies: BaseDependencies) => {
...
-				return res.send(registered);
+
+				const result: RegisterDictionaryResult = {
+					categoryId: category.id,
+					categoryName: category.name,
+					dictionary: dictionary.dictionary,
+					name: dictionary.name,
+					version: dictionary.version,
+				};
+				return res.send(result);
```
leoraba, line 50: Returning a response type `RegisterDictionaryResult` instead of any response from the service.

packages/data-provider/src/utils/categoryUtils.ts
leoraba, line 1: This file is unnecessary, it contains only one function and is not used anymore

packages/data-provider/src/services/dictionaryService.ts
```diff
@@ -1,36 +1,57 @@
...
+			const savedDictionary = await createDictionaryIfDoesNotExist(dictionaryName, version, dictionary.schemas);
+
+			// Check if Category exist
+			const foundCategory = await categoryRepo.getCategoryByName(categoryName);
+
+			if (!foundCategory) {
+				// Create a new Category
+				const newCategory: NewCategory = {
+					name: categoryName,
+					activeDictionaryId: savedDictionary.id,
```
leoraba, line 34: Creating a new Category with a reference of an Active Dictionary

packages/data-provider/src/services/dictionaryService.ts
```diff
@@ -1,36 +1,57 @@
...
-				version,
-				savedCategory,
-				dictionary.schemas,
-			);
+			const savedDictionary = await createDictionaryIfDoesNotExist(dictionaryName, version, dictionary.schemas);
+
+			// Check if Category exist
+			const foundCategory = await categoryRepo.getCategoryByName(categoryName);
+
+			if (!foundCategory) {
```
joneubank, line 30: Will suggest using the opposite check (`if (foundCategory)`) for the following reasons:
- negative logic checks are slightly trickier to understand when reading
- there is a lot of function handling when the function is not found, resulting in a lot of nested code
- the positive check returns immediately

So I would replace this with:
```ts
		if (foundCategory) {
			logger.info(LOG_MODULE, `Dictionary and Category already exists`);
			return { dictionary: savedDictionary, category: foundCategory };
		}
		// continue with code for case where foundCategory === false
```

joneubank (approved):
Looks great. One suggested change for code readability.

packages/data-provider/src/services/dictionaryService.ts
leoraba: this function handles 3 scenarios:

- foundCategory && foundCategory.activeDictionaryId === savedDictionary.id
- foundCategory && foundCategory.activeDictionaryId !== savedDictionary.id
- !foundCategory

moving the positive check first would look like this
```
if (foundCategory && foundCategory.activeDictionaryId === savedDictionary.id) {
  // code here
} else if(foundCategory && foundCategory.activeDictionaryId !== savedDictionary.id) {
   // code here
} else{ 
   // code here
} 
```

PR #71, opened 2024-07-11
add eslint and fix problems

leoraba:
Changes:
- Added eslint to this repo
- Fixed eslint problems
- Update to Lyric V0.2.2

### DIFF
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "server",
-	"version": "0.2.1",
+	"version": "0.2.2",
 	"description": "Lyric Server",
 	"engines": {
 		"node": ">=20.0.0"
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -25,7 +25,6 @@ export default tseslint.config(
 	// base config
 	{
 		rules: {
-			'@typescript-eslint/no-explicit-any': 'off',
 			'@typescript-eslint/no-unused-vars': [
 				'warn',
 				{
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -26,6 +26,12 @@ export default tseslint.config(
 	{
 		rules: {
 			'@typescript-eslint/no-explicit-any': 'off',
+			'@typescript-eslint/no-unused-vars': [
+				'warn',
+				{
+					argsIgnorePattern: '^_',
+				},
+			],
 			'simple-import-sort/imports': [
 				'warn',
 				{
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -1,0 +1,61 @@
+import js from '@eslint/js';
+import prettierPlugin from 'eslint-plugin-prettier';
+import simpleImportSortPlugin from 'eslint-plugin-simple-import-sort';
+import tseslint from 'typescript-eslint';
+
+export default tseslint.config(
+	// register all of the plugins up-front
+	{
+		plugins: {
+			['@typescript-eslint']: tseslint.plugin,
+			['simple-import-sort']: simpleImportSortPlugin,
+			['prettier']: prettierPlugin,
+		},
+	},
+
+	// config with just ignores is the replacement for `.eslintignore`
+	{
+		ignores: ['**/dist', '**/package-lock.json', '**/package.json'],
+	},
+
+	// extends ...
+	js.configs.recommended,
+	...tseslint.configs.recommended,
+
+	// base config
+	{
+		rules: {
+			'@typescript-eslint/no-explicit-any': 'off',
+			'simple-import-sort/imports': [
+				'warn',
+				{
+					groups: [
+						// Side effect imports.
+						['^\\u0000'],
+						// Node.js builtins prefixed with `node:`.
+						['^node:'],
+						// Packages.
+						// Things that start with a letter (or digit or underscore), or `@` followed by a letter.
+						['^@?\\w'],
+						// Internal packages.
+						['^(@overturebio-stack|@overture-stack)'],
+						// Absolute imports and other imports such as Vue-style `@/foo`.
+						// Anything not matched in another group.
+						['^'],
+						// Relative imports.
+						// Anything that starts with a dot.
+						['^\\.'],
+					],
+				},
+			],
+			'simple-import-sort/exports': 'error',
+			'prettier/prettier': 'error',
+		},
+		languageOptions: {
+			parserOptions: {
+				ecmaVersion: 'latest',
+				sourceType: 'module',
+			},
+		},
+	},
+);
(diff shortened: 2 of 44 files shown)


prettier.config.js
```diff
@@ -1,11 +1,9 @@
 const config = {
-	organizeImportsSkipDestructiveCodeActions: true,
 	printWidth: 120,
 	semi: true,
 	singleQuote: true,
 	trailingComma: 'all',
 	useTabs: true,
-	plugins: ['prettier-plugin-organize-imports'],
```
leoraba, line 8: Removing prettier plugin to let `eslint` organize imports as groups are defined on eslint config file

eslint.config.js
```diff
@@ -0,0 +1,61 @@
...
+					groups: [
+						// Side effect imports.
+						['^\\u0000'],
+						// Node.js builtins prefixed with `node:`.
+						['^node:'],
+						// Packages.
+						// Things that start with a letter (or digit or underscore), or `@` followed by a letter.
+						['^@?\\w'],
+						// Internal packages.
+						['^(@overturebio-stack|@overture-stack)'],
```
leoraba, line 41: Added a group for internal packages imported from overture

apps/server/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "server",
-	"version": "0.2.1",
```
joneubank, line 3: If this is just linting errors, do we need a new version/build?

package.json
```diff
@@ -17,10 +19,14 @@
...
 		"dotenv-cli": "^7.4.2",
+		"eslint": "^9.6.0",
+		"eslint-config-prettier": "^9.1.0",
+		"eslint-plugin-prettier": "^5.1.3",
+		"eslint-plugin-simple-import-sort": "^12.1.1",
 		"prettier": "^3.3.2",
-		"prettier-plugin-organize-imports": "^3.2.4",
 		"ts-node": "^10.9.2",
-		"typescript": "^5.4.5"
+		"typescript": "^5.4.5",
```
joneubank, line 28: Is this a good chance to update to TS 5.5+? Or would you like to do that on a new PR?

packages/data-provider/src/controllers/submittedDataController.ts
```diff
@@ -15,7 +16,7 @@ const controller = (dependencies: BaseDependencies) => {
 	const defaultPageSize = 20;
 	return {
 		deleteSubmittedDataBySystemId: async (
-			req: Request<{ systemId: string }, {}, any, { dryRun: string; comment: string }>,
+			req: Request<{ systemId: string }, object, any, { dryRun: string; comment: string }>,
```
joneubank, line 19: This is a good update. `{}` as a type just indicates "not null or undefined". `object` specifies something that is at least an object.

eslint.config.js
```diff
@@ -0,0 +1,61 @@
...
+	},
+
+	// extends ...
+	js.configs.recommended,
+	...tseslint.configs.recommended,
+
+	// base config
+	{
+		rules: {
+			'@typescript-eslint/no-explicit-any': 'off',
```
joneubank, line 28: What is the justification for disabling explicit any? There are bascially no cases where we want to use `any`. The suggestions provided in the rule documentation for `any` alternatives are pretty strong:

https://typescript-eslint.io/rules/no-explicit-any/

Generally, instead of any we would want to provide the minimum data type expected, or use `unknown` so any handling of that data will validate the contents.

packages/data-provider/src/utils/submissionUtils.ts
```diff
@@ -256,7 +257,7 @@ const utils = (dependencies: BaseDependencies) => {
 		parseActiveSubmissionSummaryResponse: (
 			submission: ActiveSubmissionSummaryRepository,
 		): ActiveSubmissionSummaryResponse => {
-			let dataSummary = Object.entries(submission.data).reduce(
+			const dataSummary = Object.entries(submission.data).reduce(
```
joneubank, line 260: another good catch by the linter!

joneubank (commented):
Love this addition. Put a question about one rule, and we'll wait for the changes discussed offline before approving.

eslint.config.js
```diff
@@ -0,0 +1,67 @@
...
+			'@typescript-eslint/no-unused-vars': [
+				'warn',
+				{
+					argsIgnorePattern: '^_',
```
leoraba, line 32: to ignore unused arguments starting with `_`

apps/server/package.json
leoraba: There's no need to generate a new build for this, this change shouldn't impact functionality. Will change version back to `0.2.1`

package.json
leoraba: I'd prefer to update `typescript` and other dependencies on a separate PR.

eslint.config.js
leoraba: We're using `any` mostly on Request handlers and catching errors for simplicity. Will remove this exception rule and fix to the corresponding types

leoraba:
Pull Request updated with latest comments. Ready for code review.

packages/data-provider/src/config/logger.ts
```diff
@@ -42,22 +42,22 @@ export const getLogger = (config: LoggerConfig): Logger => {
...
-		warn: (...messages: any[]) => {
+		warn: (...messages: unknown[]) => {
 			return logger.warn(log(...messages));
 		},
-		info: (...messages: any[]) => {
+		info: (...messages: unknown[]) => {
 			return logger.info(log(...messages));
 		},
-		error: (...messages: any[]) => {
+		error: (...messages: unknown[]) => {
```
joneubank, line 60: hmmmm I see this is a place where we were using `any` in a logical way. `unknown` works fine but looks weird.

packages/data-provider/src/utils/requestValidation.ts
```diff
@@ -28,7 +28,7 @@ export function validateRequest<TBody>(
 		} catch (error) {
 			if (error instanceof ZodError) {
 				const errorMessages = error.errors
-					.map((issue: any) => `${issue.path.join('.')} is ${issue.message}`)
+					.map((issue: ZodIssue) => `${issue.path.join('.')} is ${issue.message}`)
```
joneubank, line 31: Don't need type declaration here. `error.errors` is typed as `ZodIssue[]` so we can infer the type.

joneubank (approved):

packages/data-provider/src/config/logger.ts
leoraba: Will leave `unknown` as it's recommended over `any` when it's possible

Issue #73, opened 2024-07-16
Add unit test


leoraba:
Implement unit test to get an acceptable coverage of the code.

joneubank:
What is "acceptable coverage"? ;)

joneubank:
Perhaps you want to specify regions of the code that need test coverage. Create a check-list and we can update it as we implement tests?

Buwujiu:
PR merged, closing.

PR #75, opened 2024-07-18
Add unit tests to shared util functions with no dependencies

leoraba:
Changes:
- Installed testing tools (mocha and chai) to run Test suites
- Adding all tests on `test/` folder 
- Unit Testing on `utils` that doesn't need any external dependency.
- Total Coverage 8.3% of lines

Fix issue #73


packages/data-provider/test/utils/convertSqonToQuery.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+import { describe, it } from 'mocha';
+
+import { SQON } from '@overture-stack/sqon-builder';
+
+import { convertSqonToQuery, parseSQON } from '../../src/utils/convertSqonToQuery.js';
+
+function extractValues(obj: SQL | undefined, key: string) {
+	let values: SQL[] = [];
+
+	function recurse(currentObj: SQL | undefined) {
```
justincorrigible, line 12: this function needn't be defined withing extractValues, as far as I can tell.

packages/data-provider/test/utils/dictionarySchemasTestData.ts
```diff
@@ -0,0 +1,264 @@
+import { Dictionary } from '@overture-stack/lyric-data-model';
+import { SchemaDefinition, ValueType } from '@overturebio-stack/lectern-client/lib/schema-entities.js';
+
+const sampleDictionaryData: SchemaDefinition[] = [
+	{
+		name: 'study',
```
justincorrigible, line 6: is it reasonable to suggest we use a more generic test model? study and participant seem PCGL-specific, when we could be using fruits or animals to clearly exemplify the reusability of the software.

justincorrigible:
happy to see all this testing 💪 
great work, @leoraba!

packages/data-provider/test/utils/convertSqonToQuery.spec.ts
leoraba: Added types on this function as mocha tries to do typescript checks and couldn't find a way to exclude  on test files.

The purpose of this function is to convert the input SQL object from drizzle into an Array which make it easier to compare on `expect()` function

justincorrigible: to be clear, what I mean is lines 18-28 could be cut and pasted to line 8 (outside `extractValues`) and the result would be the same, minus the reinstantiation of `recurse` every time the parent function runs

packages/data-provider/test/utils/convertSqonToQuery.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+			op: 'gt',
+			content: { fieldName: 'date_of_birth', value: 197005 },
+		};
+
+		const greaterThanFilterChunk: SQLChunk[] = ["data ->> 'date_of_birth' > '197005'"];
+
+		it('should convert a json text with greater than filter into a SQON format', () => {
+			const result = parseSQON(sqonGreaterThanFilterRawInput);
+			expect(JSON.stringify(result)).to.eql(JSON.stringify(sqonGreaterThanFilterParsed));
+		});
```
joneubank, line 59: `parseSQON` appears to just be a wrapper over the SQONBuilder library... it doesn't really need test cases. If there are specific requirements you need this to perform then certainly write those. This test just checks if the outer `and` combination operator is stripped from the SQON...

packages/data-provider/test/utils/convertSqonToQuery.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+
+		it('should convert SQON with greater than filter into a database query', () => {
+			const result = convertSqonToQuery(sqonGreaterThanFilterParsed);
+			const extractedValues = extractValues(result, 'value');
+			expect(extractedValues).to.eql(greaterThanFilterChunk);
+		});
```
joneubank, line 65: Good test.

packages/data-provider/test/utils/convertSqonToQuery.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+			op: 'lt',
+			content: { fieldName: 'date_of_birth', value: 197005 },
+		};
+
+		const lessThanFilterChunk: SQLChunk[] = ["data ->> 'date_of_birth' < '197005'"];
+
+		it('should convert a json text with less than filter into a SQON format', () => {
+			const result = parseSQON(sqonLessThanFilterRawInput);
+			expect(JSON.stringify(result)).to.eql(JSON.stringify(sqonLessThanFilterParsed));
+		});
```
joneubank, line 92: Repeat comment about testing parseSQON.

packages/data-provider/test/utils/dictionarySchemasTestData.ts
```diff
@@ -0,0 +1,213 @@
+import { Dictionary } from '@overture-stack/lyric-data-model';
+import { SchemaDefinition, ValueType } from '@overturebio-stack/lectern-client/lib/schema-entities.js';
+
+const sampleDictionaryData: SchemaDefinition[] = [
```
joneubank, line 4: You may want to create a dir of `test/fixtures` that includes test dictionaries

joneubank: This dictionary could be called `dictionarySportStats`

packages/data-provider/test/utils/dictionarySchemasTestData.ts
```diff
@@ -0,0 +1,213 @@
...
+export const sampleDictionary: Dictionary = {
+	id: 1,
+	name: 'test dictionary',
+	version: '1.0.0',
+	dictionary: sampleDictionaryData,
+	createdAt: new Date(),
+	createdBy: '',
+};
+
+export const sampleChildrenNodes = {
```
joneubank, line 201: `dictionarySportStatsNodeGraph` or something more descriptive. We will end up with many `samples` as we develop.

packages/data-provider/test/utils/dictionarySchemasTestData.ts
```diff
@@ -0,0 +1,213 @@
...
+		{ schemaName: 'game', fieldName: 'sport_id', parent: { schemaName: 'sport', fieldName: 'sport_id' } },
+	],
+	team: [
+		{
+			schemaName: 'player',
+			fieldName: 'team_id',
+			parent: { schemaName: 'team', fieldName: 'team_id' },
+		},
+	],
+};
```
joneubank, line 213: `as const` so no tests try to modify our test fixture.

joneubank: Repeat this comment for all the test fixtures.

packages/data-provider/test/utils/fileUtils.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { expect } from 'chai';
+import fs from 'fs';
+import { describe, it } from 'mocha';
+import sinon from 'sinon';
+
+const fsPromises = fs.promises;
+
+import { tsvToJson } from '../../src/utils/fileUtils.js';
+
+describe('File Utils tests', () => {
```
joneubank, line 10: are there any failure cases we should test? behaviour we expect when the file is not a tsv? empty file behaviour? your example only has a single record, should we test one with multiple records? a header row and no records?

packages/data-provider/test/utils/formatUtils.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+describe('Format Utils tests', () => {
+	describe('Validate if input is a valid ID number', () => {
+		it('should return true if input is a valid number', () => {
+			const validNumber = 1;
+			const response = isValidIdNumber(validNumber);
+			expect(response).to.be.true;
+		});
+
+		it('should return false if input is a not a valid number', () => {
+			const invalidNumber = 'one';
```
joneubank, line 15: Are there other edge cases to test? empty values, numeric strings like `'1'`, `Infinity`/`NaN`, null/undefined, negative numbers, `0`, numbers beyond max int...

packages/data-provider/test/utils/formatUtils.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+			const response = isValidIdNumber(invalidNumber);
+			expect(response).to.be.false;
+		});
+	});
+
+	describe('Extract unique characters on a String', () => {
+		it('should return unique characters on a string', () => {
+			const duplicateCharactersString = 'AAABBC';
+			const uniqueCharactersString = 'ABC';
+			const response = uniqueCharacters(duplicateCharactersString);
```
joneubank, line 25: What do we use this for? Seems cool...

packages/data-provider/test/utils/formatUtils.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+			expect(response).to.eql(uniqueCharactersString);
+		});
+	});
+
+	describe('Validate if input string is empty or not', () => {
+		it('should return true if string is empty', () => {
+			const emptyString = '';
```
joneubank, line 32: What about the string `'       '`

packages/data-provider/test/utils/formatUtils.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+			const response = isEmptyString(emptyString);
+			expect(response).to.be.true;
+		});
+
+		it('should return false if string is not empty', () => {
+			const emptyString = 'test';
```
joneubank, line 38: other edge case? null/undefined/false/0/NaN/[]

packages/data-provider/test/utils/convertSqonToQuery.spec.ts
```diff
@@ -0,0 +1,250 @@
...
+			}
+			Object.values(currentObj).forEach(recurse);
+		}
+	}
+
+	recurse(obj);
+	return values;
+}
+
+describe('Test SQON functions', () => {
```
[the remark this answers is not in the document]
leoraba: renamed as suggested

packages/data-provider/test/utils/convertSqonToQuery.spec.ts
leoraba: `parseSQON` is a wrapper of a SQONBuilder function and we're not adding any logic. Will remove this test .

packages/data-provider/test/utils/dictionarySchemasTestData.ts
leoraba: Moved this file to `fixtures/` folder and changed it's variable names

packages/data-provider/test/utils/dictionarySchemasTestData.ts
leoraba: renamed const name to `dictionarySportStatsNodeGraph `

packages/data-provider/test/utils/dictionarySchemasTestData.ts
leoraba: Good catch. Added const assertion.

packages/data-provider/test/utils/formatUtils.spec.ts
leoraba: we use this function to generate system IDs using only unique characters when user sets a custom alphabet

packages/data-provider/test/utils/fileUtils.spec.ts
leoraba: adding more tests to cover more edge scenarios

README.md
```diff
@@ -26,7 +26,7 @@ The structure of this monorepo is app centric having `apps/` folder to keep depl
 
 - [PNPM](https://pnpm.io/) Project manager
 - [Node.js](https://nodejs.org/en) Runtime environment (v20 or higher)
-- [VS Code](https://code.visualstudio.com/) As recommended code editor
+- [VS Code](https://code.visualstudio.com/) As recommended code editor. Plugins recommended: ESLint, Prettier - Code formatter, Mocha Test Explorer, Monorepo Workspace
```
joneubank, line 29: 👍

joneubank (approved):
Great stuff, thanks.

PR #80, opened 2024-07-23
Add `audit` endpoint to retrieve change history records for submitted data

leoraba:
Implementation for **GET** `/audit/category/:categoryId/organization/:organization` endpoint.

Includes:
- Added Swagger doc to describe Audit endpoint
- Optional Filter options (query params): *entityName, eventType, systemId* 
- Pagination support
- Unit testing for audit utils.
- Using a Request hander function to validate input request using Zod schemas


apps/server/src/server.ts
```diff
@@ -22,6 +22,7 @@ app.use(helmet());
 app.use('/ping', pingRouter);
 
 // Lyric Routes
+app.use('/audit', lyricProvider.routers.audit);
```
leoraba, line 25: Adding new endpoint to express server

packages/data-provider/src/utils/auditUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Convert a value string into it's Audit event type if it matches.
+ * Otherwise it returns `undefined`
+ * @param {string} value
+ * @returns {AuditAction | undefined}
+ */
+export const convertToAuditEvent = (value: string): AuditAction | undefined => {
+	if (isAuditEventValid(value)) {
+		if (value.toUpperCase() === AUDIT_ACTION.UPDATE.toString()) return AUDIT_ACTION.UPDATE;
+		else if (value.toUpperCase() === AUDIT_ACTION.DELETE.toString()) return AUDIT_ACTION.DELETE;
```
joneubank, line 27: No need for these toString() calls. AUDIT_ACTION.DELETE is already a string.

packages/data-provider/src/utils/auditUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Otherwise returns `false`
+ * @param {unknown} value
+ * @returns {boolean}
+ */
+export const isAuditEventValid = (value: unknown): boolean => {
+	return (
+		typeof value === 'string' &&
+		[AUDIT_ACTION.DELETE.toString(), AUDIT_ACTION.UPDATE.toString()].includes(value.toUpperCase())
+	);
+};
```
joneubank, line 16: These `toString()` calls are distracting, and only required because of the `includes` interface being strict to the array const types. Additionally, there is no guarantee that the AUDIT_ACTION constants are all upper case, so we should probably enforce that with a `toUpperCase` map.  

Consider instead:
```suggestion
const upperCaseAuditActions: string[] = Object.values(AUDIT_ACTION).map((value) => value.toUpperCase());
export const isAuditEventValid = (value: unknown): boolean => 
	typeof value === 'string' && upperCaseAuditActions.includes(value.toUpperCase());
```

I was going to recommend using a zod enum since we are writing our own matcher... but I can't figure out how to do case insensitive matching using zod enum.

packages/data-provider/src/utils/types.ts
```diff
@@ -30,6 +30,60 @@ export const AUDIT_ACTION = {
 } as const;
 export type AuditAction = ObjectValues<typeof AUDIT_ACTION>;
 
+/**
+ * Audit Raw Data from Repository
+ */
+export type AuditDataRepository = {
```
joneubank, line 36: Being picky... This isnt the type for the Repository, its the type for a Repository record. In my java days they would call this the DAO (data access object). Would you consider the name `AuditRepositoryRecord`

packages/data-provider/test/utils/auditUtils.spec.ts
```diff
@@ -0,0 +1,71 @@
+import { expect } from 'chai';
+import { describe, it } from 'mocha';
+
+import { convertToAuditEvent, isAuditEventValid, parseAuditRecords } from '../../src/utils/auditUtils.js';
+import { AuditDataRepository, AuditDataResponse } from '../../src/utils/types.js';
+
+describe('Audit functions test', () => {
```
joneubank, line 7: `test` is redundant. maybe just `describe('Audit utils')` since we are in a file called auditUtils.spec and testing a file called `auditUtils` ;)

joneubank (changes requested):
Minor changes recommended.

packages/data-provider/src/controllers/auditController.ts
```diff
@@ -0,0 +1,111 @@
...
+					throw new BadRequest('Invalid `pageSize` parameter');
+				}
+
+				if (!isEmptyString(startDate) && !isValidDateFormat(startDate)) {
+					throw new BadRequest('Invalid `startDate` parameter');
+				}
+
+				if (!isEmptyString(endDate) && !isValidDateFormat(endDate)) {
+					throw new BadRequest('Invalid `endDate` parameter');
+				}
```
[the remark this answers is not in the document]
joneubank: You could also create a `PaginationQuery` schema that can be reused in many places.

packages/data-provider/test/utils/auditUtils.spec.ts
leoraba: renamed test description as `Audit utils`

packages/data-provider/src/utils/types.ts
leoraba: renamed to `AuditRepositoryRecord `

packages/data-provider/src/utils/auditUtils.ts
leoraba: Updated this function to check if the audit event is valid using Zod safeParse.

packages/data-provider/src/utils/auditUtils.ts
leoraba: Updated this function to check using Zod instead

packages/data-provider/src/controllers/auditController.ts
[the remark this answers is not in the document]
leoraba: Implemented a Request hander to validate request input using Zod schemas

leoraba:
Updated this PR to follow comments and suggestions. This is now ready for code review.

packages/data-provider/src/utils/requestValidation.ts
```diff
@@ -1,29 +1,38 @@
...
-	schema: RequestValidation<TBody>,
-	handler: RequestHandler<ParamsDictionary, unknown, TBody>,
-): RequestHandler {
+export function validateRequest<TBody, TQuery, TParams>(
+	schema: RequestValidation<TBody, TQuery, TParams>,
+	handler: RequestHandler<TParams, unknown, TBody, TQuery>,
+): RequestHandler<TParams, unknown, TBody, TQuery> {
 	const LOG_MODULE = 'REQUEST_VALIDATION';
-	return async (req: Request, res: Response, next: NextFunction) => {
+	return async (req, res, next) => {
```
joneubank, line 22: :+1:

packages/data-provider/src/utils/auditUtils.ts
```diff
@@ -1,19 +1,31 @@
...
+
+/**
+ * This function takes an array of values and casts it to a tuple type where the first element is of the same type as the array, and the rest are of the same type as well.
+ */
+const zodEnum = <T>(arr: T[]): [T, ...T[]] => arr as [T, ...T[]];
+
+/**
+ * Create Zod enum schema from uppercase enum values
+ */
+const auditActionSchema = z.enum(zodEnum(upperCaseAuditActions));
```
joneubank, line 19: This is all very round about way of defining an enum with a zod schema. Just make the original declaration of AUDIT_ACTION using a zod.enum. This will also make the type assertion in the return statement below unnecessary.

packages/data-provider/src/utils/schemas.ts
```diff
@@ -1,27 +1,317 @@
...
+import { isAuditEventValid } from './auditUtils.js';
+import { parseSQON } from './convertSqonToQuery.js';
+import { isValidDateFormat, isValidIdNumber } from './formatUtils.js';
 import { RequestValidation } from './requestValidation.js';
 
-export const uploadCustomFile = z
-	.custom<Express.Multer.File[]>()
-	.refine((files) => files?.length > 0, 'required as a .tsv format')
-	.refine((files) => files.every((file) => file.originalname.match(/.*\.tsv$/)), 'required as a .tsv format');
+// Zod Schemas
```
joneubank, line 12: comment doesn't really communicate anything... this whole file is zod schema definitions....

joneubank (changes requested):
Simplify that one enum so we don't need so many ways of defining the same types and don't need the type assertion.

packages/data-provider/src/utils/auditUtils.ts
leoraba: Used another approach to avoid impact other files that are using original AUDIT_ACTION. 
Assigned a Zod nativeEnum with AUDIT_ACTION and transform it to uppercase values maintaining the type.

packages/data-provider/src/utils/schemas.ts
leoraba: Removed unnecessary comment

leoraba:
PR updated. Ready for code review

packages/data-provider/src/utils/auditUtils.ts
```diff
@@ -4,19 +4,9 @@ import { z } from 'zod';
...
-/**
- * This function takes an array of values and casts it to a tuple type where the first element is of the same type as the array, and the rest are of the same type as well.
- */
-const zodEnum = <T>(arr: T[]): [T, ...T[]] => arr as [T, ...T[]];
-
-/**
- * Create Zod enum schema from uppercase enum values
- */
-const auditActionSchema = z.enum(zodEnum(upperCaseAuditActions));
+const auditActionSchema = z.nativeEnum(AUDIT_ACTION).transform((value) => value.toUpperCase());
```
joneubank, line 9: You can remove the toUpperCase here to make sure that `auditActionSchema` parses into the enum string literals. You perform the `toUpperCase` conversion in `isAuditEventValid` and `convertToAuditEvent` before parsing anyways. My objective here is to be able to remove the `as AUDIT_ACTION` on the `convertToAuditEvent` return statement.

joneubank (changes requested):

packages/data-provider/src/utils/auditUtils.ts
leoraba: converted original `AUDIT_ACTION` to zod enum to simplify validations

packages/data-provider/src/utils/auditUtils.ts
```diff
@@ -24,7 +19,7 @@ export const isAuditEventValid = (value: unknown): boolean =>
  * @returns {AuditAction | undefined}
  */
 export const convertToAuditEvent = (value: string): AuditAction | undefined => {
-	const parseResult = auditActionSchema.safeParse(value.toUpperCase());
+	const parseResult = AUDIT_ACTION.transform((action) => action.toUpperCase()).safeParse(value.toUpperCase());
```
joneubank, line 22: you are transforming the value before sending it to the parser, you don't need to add the transform!

leoraba: removed transform function as we're already passing the uppercase value

joneubank (approved):
LGTM!

PR #83, opened 2024-07-26
Add Entity Filter to Get Submitted Data Endpoint

leoraba:
Add `entityName` Filter as a query parameter to GET data endpoints:
1. GET /data/category/{categoryId}
2. GET /data/category/{categoryId}/organization/{organization}
3. GET /data/category/{categoryId}/organization/{organization}/query:

Include:
- Accepts a string to filter query by entityName. Separate multiple values with commas
- Added query param on Swagger docs


packages/data-provider/src/controllers/submittedDataController.ts
```diff
@@ -58,12 +58,15 @@ const controller = (dependencies: BaseDependencies) => {
 		},
 
 		getSubmittedDataByCategory: async (
-			req: Request<{ categoryId: string }, object, object, { pageSize: string; page: string }>,
+			req: Request<{ categoryId: string }, object, object, { entityName: string; pageSize: string; page: string }>,
```
leoraba, line 61: typing new query param `entityName` on the request

packages/data-provider/src/repository/submittedRepository.ts
```diff
@@ -47,6 +47,20 @@ const repository = (dependencies: BaseDependencies) => {
 		systemId: true,
 	};
 
+	/**
+	 * Build a SQL object to search submitted data by entity Name
+	 * @param {string[] | undefined} entityNameArray
+	 * @returns {SQL<unknown> | undefined}
+	 */
+	const filterByEntityNameArray = (entityNameArray: string[] | undefined): SQL<unknown> | undefined => {
```
leoraba, line 55: function to convert an array of string into Drizzle SQL query

packages/data-provider/src/utils/schemas.ts
```diff
@@ -290,26 +290,38 @@ export const dataDeleteBySystemIdRequestSchema: RequestValidation<
...
+	entityName?: string;
+}
+
+export const dataGetByCategoryRequestSchema: RequestValidation<object, dataQueryParams, categoryPathParams> = {
+	query: z
+		.object({
+			entityName: entityNameSchema.optional(),
+		})
+		.merge(paginationQuerySchema),
 	pathParams: categoryPathParamsSchema,
```
leoraba, line 303: validate Request query and path params using Zod schema

packages/data-provider/src/utils/formatUtils.ts
```diff
@@ -8,6 +8,10 @@ export const isArrayWithValues = (value: unknown) => {
 	return Array.isArray(value) && value.length > 0 && value.some((x) => !!x);
 };
 
+export const splitString = (value: string, separator: string) => {
+	return value.split(',');
```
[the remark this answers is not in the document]
leoraba: Thanks for pointing this out. Turns out this function is not used as expected. 
1) it receives a second argument to pass a separator and is not being used.
2) The function is not even called. the supposed caller is using it's own logic to split strings.

Solution: 
1) To use the argument to set the separator
2) Call this function whenever it's required to split strings, where we can set any separator to use, for this case we can use ";"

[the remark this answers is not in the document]
leoraba: code changed to solve previous comment

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -16,6 +17,10 @@ const PAGINATION_ERROR_MESSAGES = {
 	NO_DATA_FOUND: 'No Submitted data found',
 } as const;
 
+// A constant representing a list separator using the pipe character (|),
+// which cannot be used in entity names due to its illegality.
+const entitySeparator = '|' as const;
```
justincorrigible, line 22: For a future PR (this one is fine as is), I'd make this a fallback value within the function, while still allowing to pass custom separators

justincorrigible (approved):
Squasharoooo 🦘

apps/server/swagger/data-api.yml
```diff
@@ -10,6 +10,10 @@
         schema:
           type: string
         description: ID of the category
+      - name: entityName
+        description: Filter by entity name. Separate multiple values with pipe character (|)
```
joneubank, line 14: Why `|` instead of `,`? using commas in query parameters is standard. We can also update the Lectern schema to remove some more special characters from schema names (no `,` or `|`)

packages/data-provider/src/repository/submittedRepository.ts
```diff
@@ -130,11 +144,15 @@ const repository = (dependencies: BaseDependencies) => {
...
+			filter?: { entityNames?: string[] },
 		): Promise<SubmittedDataResponse[]> => {
 			const { page, pageSize } = paginationOptions;
+
+			const filterEntityNameSql = filterByEntityNameArray(filter?.entityNames);
+
 			try {
 				return await db.query.submittedData.findMany({
-					where: eq(submittedData.dictionaryCategoryId, categoryId),
+					where: and(eq(submittedData.dictionaryCategoryId, categoryId), filterEntityNameSql),
```
joneubank, line 155: how does this work when `filterEntityNameSql` is `undefined`? any issues? this will happen when there are no entityName query params provided.

joneubank: Based on some other uses of the drizzle `and()` it looks like `undefined` is commonly used like this, so no worries here :+1:

apps/server/swagger/data-api.yml
joneubank: Maybe its even better to use the HTTP query param pattern where an array is created when the query parameter is used multiple times:

`GET /path/to/thing?entityName=donor&entityName=specimen&entityName=sample`

This is actually how Swagger will build arrays when you specify an array query param, and express will parse this as a single array. This also means there is no custom parsing needed to split a value into an array, which has led to comments in this review about how best to standardize that (from Justin below and me above here).

justincorrigible: While I realise this is just an example, there are multiple valid formats to build http array query params, without clear consensus on an approach. Enforcing one would limit third party integration, no?

On the other hand, the pipe here is meant to just be an alternative default separator, for what in my initial comment was "don't hardcode comma as the only possible option", since CSVs allow commas within a value if you wrap the value in single/double quotes.
I'm currently dealing with that in the wastewater implementation.

packages/data-provider/src/repository/submittedRepository.ts
leoraba: there is no issues when filterEntityNameSql is `undefined` Drizzle handles it with no issues.

apps/server/swagger/data-api.yml
leoraba: This makes more sense. instead of handling special character separators ourselves will use the standard query param array format.

leoraba:
PR updated to incorporate the feedback.

packages/data-provider/src/utils/formatUtils.ts
```diff
@@ -62,3 +62,15 @@ export function isValidDateFormat(value: string): boolean {
...
+ * Otherwise, it returns the input as-is.
+ * @param {string | string[] | undefined} value
+ * @returns {string[] | undefined}
+ */
+export function stringToArray(value: string | string[] | undefined): string[] | undefined {
+	if (typeof value === 'string') {
+		return [value];
+	}
+	return value;
+}
```
joneubank, line 76: This works and is a good solution. I have similar utility I use throughout lectern that I thought I'd share as a general solution: 

```ts
/**
 * Ensure a value is wrapped in an array.
 *
 * If passed an array, return it without change. If passed a single item, wrap it in an array.
 * @param val an item or array
 * @return an array
 */
export const asArray = <T>(val: T | T[]): T[] => (Array.isArray(val) ? val : [val]);
```

https://github.com/overture-stack/lectern/blob/develop/packages/dictionary/src/utils/typeUtils.ts#L20C1-L27C85

joneubank (approved):

packages/data-provider/src/utils/formatUtils.ts
leoraba: Thanks for sharing this solution. Will use `asArray` function as I see it's more generic and will always return an array `(string | undefined)[]`

justincorrigible: While the implementation for both your functions is basically the same 👍 (a condition with the same two outcomes), the name for both of your functions would lead me to believe the functions do something entirely different. Naming things is hard, for sure... but hear me out:

- `String to Array` makes me think you'd parse `blah, blah, blah` into an array with the three "blah" strings.
- `asArray` evokes some sort of type assertion.
- ultimately neither name indicates "I'll wrap your value in an array, unless the value is already an array"

Not sure whether suggesting `arrayFromValue` or `valueToArray` makes it better, but perhaps addressing possible "expectations" from what the function may do (e.g. parsing a string into an array if possible, rather than just wrapping it) may help later.

PR #84, opened 2024-07-30
Swagger doc changes

leoraba:
**Changes include:**
- To extend Swagger description of input fields
- To provide SQON examples


packages/data-provider/test/utils/formatUtils.spec.ts
```diff
@@ -175,4 +181,36 @@ describe('Format Utils', () => {
 			expect(response).to.be.true;
 		});
 	});
+
+	describe('Validate if input is a valid date', () => {
+		it('should return true if input is a YYYY-MM-DD string date format', () => {
+			const input = '2000-01-01';
+			const response = isValidDateFormat(input);
+			expect(response).to.be.true;
```
joneubank, line 189: just fyi `.to.be` don't actually do anything, its just for readability. Can also be written as `expect(value).true`

joneubank (approved):

PR #93, opened 2024-08-14
:wastebasket: Delete Submitted Data using an active Submission

leoraba:
Description of the changes:
- `DELETE data/{systemId}` endpoint creates an Active Submission with the record(s) to be deleted
- Active Submission operations(GET, POST, DELETE) now can handle mixed Submission where are records to be **inserted** or **deleted**.  
- `Upload`, `Delete` and `Commit` Submission endpoints perform a cross schema validation.
- Swagger doc changes

### DIFF
--- a/apps/server/swagger/data-api.yml
+++ b/apps/server/swagger/data-api.yml
@@ -2,7 +2,7 @@
   get:
     summary: Retrieve Submitted Data for a specific Category
     tags:
-      - data
+      - Data
     parameters:
       - name: categoryId
         in: path
@@ -53,7 +53,7 @@
   get:
     summary: Retrieve Submitted Data for a specific Category and Organization
     tags:
-      - data
+      - Data
     parameters:
       - name: categoryId
         in: path
@@ -110,7 +110,7 @@
   post:
     summary: Retrieve Submitted Data filtered by search parameters provided in the request body
     tags:
-      - data
+      - Data
     parameters:
       - name: categoryId
         in: path
@@ -226,7 +226,7 @@
   delete:
     summary: Delete Submitted Data by System ID. Returns an Active Submission with the records to be deleted
     tags:
-      - admin
+      - Admin
     parameters:
       - name: systemId
         in: path
--- a/apps/server/swagger/data-api.yml
+++ b/apps/server/swagger/data-api.yml
@@ -224,7 +224,7 @@
 
 /data/{systemId}:
   delete:
-    summary: Delete Submitted Data by System ID
+    summary: Delete Submitted Data by System ID. Returns an Active Submission with the records to be deleted
     tags:
       - admin
     parameters:
@@ -233,26 +233,13 @@
         required: true
         schema:
           type: string
-      - name: dryRun
-        description: If true won't execute the deletion, but will return what will be deleted, default is false.
-        in: query
-        required: false
-        schema:
-          default: false
-          type: boolean
-      - name: comment
-        description: Comment explaining the reason for deletion.
-        in: query
-        required: false
-        schema:
-          type: string
     responses:
       200:
-        description: Submitted Data cleared successfully. Returns the delete Submitted Data
+        description: Returns an Active Submission with the records to be deleted
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/GetSubmittedDataResult'
+              $ref: '#/components/schemas/DeleteDataActiveSubmission'
       400:
         $ref: '#/components/responses/BadRequest'
       401:
--- a/apps/server/swagger/submission-api.yml
+++ b/apps/server/swagger/submission-api.yml
@@ -56,7 +56,7 @@
       503:
         $ref: '#/components/responses/ServiceUnavailableError'
 
-/submission/{submissionId}/{entityName}:
+/submission/{submissionId}/{actionType}:
   delete:
     summary: Clear Active Submission by entity name
     tags:
@@ -66,10 +66,25 @@
         in: path
         type: string
         required: true
-      - name: entityName
+        description: The ID of the Submission
+      - name: actionType
         in: path
+        required: true
+        schema:
+          type: string
+          enum: [inserts, updates, deletes]
+        description: Optional query parameter to specify the type of submission to delete
+      - name: entityName
+        in: query
         type: string
         required: true
+        description: The name of the entity
+      - name: index
+        in: query
+        required: false
+        schema:
+          type: integer
+        description: An optional query parameter used to specify the index of the item within the submission type to be deleted. <br />If not provided all the items within the submission type will be deleted.
     responses:
       200:
         description: Submission cleared successfully. Returns the current Active Submission
--- a/apps/server/swagger/submission-api.yml
+++ b/apps/server/swagger/submission-api.yml
@@ -73,7 +73,7 @@
         schema:
           type: string
           enum: [inserts, updates, deletes]
-        description: Optional query parameter to specify the type of submission to delete
+        description: Parameter to specify the type of record to remove from the Submission. Must be one of `inserts`, `updates`, or `deletes` (case insensitive)
       - name: entityName
         in: query
         type: string
--- a/apps/server/swagger/submission-api.yml
+++ b/apps/server/swagger/submission-api.yml
@@ -4,7 +4,7 @@
   get:
     summary: Get active submission by ID
     tags:
-      - submission
+      - Submission
     parameters:
       - name: submissionId
         in: path
@@ -30,7 +30,7 @@
   delete:
     summary: Clear an Active Submission by ID
     tags:
-      - submission
+      - Submission
     parameters:
       - name: submissionId
         in: path
@@ -60,7 +60,7 @@
   delete:
     summary: Clear Active Submission by entity name
     tags:
-      - submission
+      - Submission
     parameters:
       - name: submissionId
         in: path
@@ -94,7 +94,7 @@
   get:
     summary: Retrieve the Active Submissions for a category in this user session
     tags:
-      - submission
+      - Submission
     parameters:
       - name: categoryId
         in: path
@@ -121,7 +121,7 @@
   get:
     summary: Retrieve the Active Submission for a category and organization in this user session
     tags:
-      - submission
+      - Submission
     parameters:
       - name: categoryId
         in: path
@@ -152,7 +152,7 @@
   post:
     summary: Upload submission files
     tags:
-      - submission
+      - Submission
     consumes:
       - multipart/form-data
     parameters:
@@ -193,7 +193,7 @@
   post:
     summary: Commit Active Submission
     tags:
-      - submission
+      - Submission
     parameters:
       - name: categoryId
         in: path
--- a/packages/data-model/docs/submissionCommit.md
+++ b/packages/data-model/docs/submissionCommit.md
@@ -1,6 +1,6 @@
 ```mermaid
 ---
-title: State Diagram Submission Statuses
+title: Commit Submission Sequence Diagram
 ---
 sequenceDiagram
     actor User
--- a/packages/data-model/docs/submissionCommit.md
+++ b/packages/data-model/docs/submissionCommit.md
@@ -9,27 +9,22 @@ sequenceDiagram
     Lyric->>+DB: get Submission by submissionId
     DB-->>-Lyric: Return Submission
 
-    Lyric->>Lyric: Check if submission exists
-    alt
-    
-        Lyric->>User: 400 Bad Request:<br />Submission '$submissionId}' not found
+    alt No Submission found
+        Lyric->>User: 400 Bad Request:<br />Submission '${submissionId}' not found
     end
 
-    Lyric->>Lyric: Check submission category matches `categoryId` param
-    alt
+    alt Found Submission category doesn't match the `categoryId` in the request parameter.
         Lyric->>User: 400 Bad Request:<br />Category ID provided does not match the category for the Submission
     end
 
-    Lyric->>Lyric: Check submission status is `valid`
-    alt
-        Lyric->>User: 407 State Conflict:<br />Only Submissions with statuses "OPEN", "VALID", "INVALID" can be deleted
+    alt Found Submission status is not `valid`
+        Lyric->>User: 407 State Conflict:<br />Submission does not have status VALID and cannot be committed
     end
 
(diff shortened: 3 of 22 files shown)


apps/server/swagger/data-api.yml
```diff
@@ -224,35 +224,22 @@
...
-        required: false
-        schema:
-          default: false
-          type: boolean
-      - name: comment
-        description: Comment explaining the reason for deletion.
-        in: query
-        required: false
-        schema:
-          type: string
```
leoraba, line 248: Removed both `dryRun` and `comment` fields. These are no longer required as Deletion is handled by active submission

apps/server/swagger/submission-api.yml
```diff
@@ -56,20 +56,35 @@
       503:
         $ref: '#/components/responses/ServiceUnavailableError'
 
-/submission/{submissionId}/{entityName}:
+/submission/{submissionId}/{actionType}:
```
leoraba, line 59: Changed this operation to receive new params `actionType,  `entityName` and `index` to delete specific items within the Active Submission

packages/data-model/docs/submissionCommit.md
leoraba, line 1: Additional sequence diagram to describe  Commit Submission logic

packages/data-model/src/models/submissions.ts
```diff
@@ -25,6 +25,10 @@ export type SubmissionUpdateData = {
 
 export type SubmissionDeleteData = {
 	systemId: string;
+	data: DataRecord;
+	entityName: string;
+	isValid: boolean;
+	organization: string;
 };
```
leoraba, line 32: Data to be stored on an Active Submission on items for deletion

packages/data-provider/src/repository/submittedRepository.ts
```diff
@@ -64,11 +66,11 @@ const repository = (dependencies: BaseDependencies) => {
...
 	return {
-		delete: async (data: SubmittedData, comment: string, userName: string) => {
-			const deletedRecord = await db.delete(submittedData).where(eq(submittedData.id, data.id));
+		deleteBySystemId: async (systemId: string, userName: string) => {
+			const deletedRecord = await db.delete(submittedData).where(eq(submittedData.systemId, systemId)).returning();
 
-			if (features?.audit?.enabled) {
-				await auditDeleteSubmittedData(data, comment, userName);
+			if (features?.audit?.enabled && deletedRecord) {
+				await auditDeleteSubmittedData(deletedRecord, userName);
```
leoraba, line 73: Delete records that are marked for deletion under the Submission. Audit records if option is enabled

packages/data-provider/src/utils/schemas.ts
```diff
@@ -242,18 +250,27 @@ export const submissionDeleteRequestSchema: RequestValidation<object, ParsedQs,
...
+	submissionDeleteEntityNameQueryParams,
 	submissionDeleteEntityNameParams
 > = {
+	query: z.object({
+		entityName: entityNameSchema,
+		index: indexIntegerSchema.optional(),
+	}),
 	pathParams: z.object({
+		actionType: submissionActionTypeSchema,
 		submissionId: submissionIdSchema,
```
leoraba, line 273: query and path params  added on Delete item on Submission endpoint

packages/data-provider/src/services/submissionService.ts
leoraba, line 1: Most of this is just a refactoring to move out functions.

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -195,25 +195,35 @@ const service = (dependencies: BaseDependencies) => {
...
+			// Get Active Submission or Open a new one
+			const activeSubmission = await getOrCreateActiveSubmission({
+				categoryId: foundRecordToDelete.dictionaryCategoryId,
+				userName,
+				organization: foundRecordToDelete.organization,
+			});
 
-			return mapRecordsSubmittedDataResponse(recordsToUpdate);
+			// Merge current Active Submission delete entities
+			const mergedSubmissionDeletes = mergeRecords(activeSubmission.data.deletes, recordsToDeleteMap);
```
leoraba, line 212: Just add items marked for deletion on the Submission instead of delete them right away

apps/server/swagger/submission-api.yml
```diff
@@ -56,20 +56,35 @@
...
-      - submission
+      - Submission
     parameters:
       - name: submissionId
         in: path
         type: string
         required: true
-      - name: entityName
+        description: The ID of the Submission
+      - name: actionType
```
justincorrigible, line 70: if this is listed here, does that make it an additional query param? e.g. `...?actionType=blah`
... wouldn't that be redundant to the endpoint's definition in line 59?

joneubank: It is listed as `in: path`

justincorrigible: to confirm, does that mean in that third segment of the path, I could provide either of those value types, and the app will somehow know how to identify it?

packages/data-provider/src/controllers/submissionController.ts
```diff
@@ -58,14 +61,24 @@ const controller = (dependencies: BaseDependencies) => {
 		deleteEntityName: validateRequest(submissionDeleteEntityNameRequestSchema, async (req, res, next) => {
 			try {
 				const submissionId = Number(req.params.submissionId);
-				const entityName = req.params.entityName;
+				const actionType = SUBMISSION_ACTION_TYPE.parse(req.params.actionType.toUpperCase());
+
+				const entityName = req.query.entityName;
+				const index = req.query.index ? parseInt(req.query.index) : undefined;
```
justincorrigible, line 67: @joneubank may disagree, but I think this is a good place to use `Null` as an explicit description of the value being missing "on purpose".
Setting something `= undefined` smells.

apps/server/swagger/data-api.yml
```diff
@@ -224,35 +224,22 @@
 
 /data/{systemId}:
   delete:
-    summary: Delete Submitted Data by System ID
+    summary: Delete Submitted Data by System ID. Returns an Active Submission with the records to be deleted
```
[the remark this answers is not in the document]
leoraba: This endpoint doesn't return any errors.  It returns only a `submission ID` and and an array of records to be marked for deletion on the Submission. 
After that you can use the returned submission ID to review any errors associated with the Submission.

apps/server/swagger/submission-api.yml
leoraba: This DELETE endpoint provides the ability to "clean" the Active Submission by choosing the specific item to remove from the Submission. Submission contains a group of staged changes separated on:
- New Uploads(`inserts`),
- Edit Submitted Data(`updates`) and 
- Delete Submitted Data(`deletes`)

Example URL: /submission/102/inserts?entityName=sample&index=0

Where **actionType** is path param to specify the type of record to remove from the Submission. Must be one of: `inserts`, `updates`, or `deletes` (case insensitive)

Thanks for pointing this out. Will fix the description of actionType that is not correct. **

packages/data-provider/src/services/submittedDataService.ts
```diff
@@ -165,28 +167,26 @@ const service = (dependencies: BaseDependencies) => {
...
 			const { getDictionaryById } = dictionaryRepo;
-			const { mapRecordsSubmittedDataResponse } = submittedUtils(dependencies);
+			const { getOrCreateActiveSubmission, mergeRecords } = submissionUtils(dependencies);
+			const { mapRecordsSubmittedDataResponse, transformmSubmittedDataToSubmissionDeleteData } =
+				submittedUtils(dependencies);
+			const { performDataValidation } = submissionService(dependencies);
 
 			// get SubmittedData by SystemId
-			const submittedData = await getSubmittedDataBySystemId(systemId);
+			const foundRecordToDelete = await getSubmittedDataBySystemId(systemId);
```
justincorrigible, line 180: Great renaming choice!

packages/data-provider/src/utils/submissionUtils.ts
```diff
@@ -230,6 +232,23 @@ const utils = (dependencies: BaseDependencies) => {
...
+		 * @param record1 The first `Record<string, T[]>` object. If `undefined`, it is treated as an empty record.
+		 * @param record2 The second `Record<string, T[]>` object. If `undefined`, it is treated as an empty record.
+		 * @returns
+		 */
+		mergeRecords: <T>(
+			record1: Record<string, T[]> | undefined,
+			record2: Record<string, T[] | undefined>,
+		): Record<string, T[]> => {
+			return Object.keys({ ...record1, ...record2 }).reduce<Record<string, T[]>>((acc, key) => {
+				acc[key] = [...(record1?.[key] || []), ...(record2?.[key] || [])];
```
justincorrigible, line 247: would the items in the merged array need to be sorted somehow?

e.g. current: `[A,C] + [B, D] = [A, C, B, D]` when it could result in `[A, B, C, D]` to reduce search complexity later on. 

please dismiss if this is "premature optimisation".

justincorrigible: oh! in that case this may be more memory efficient and easier to read/reason, no?
```suggestion
				acc[key] = (record1?.[key] || []).concat(record2?.[key] || []);
```

leoraba: Thanks for the suggestion. Code fixed.

packages/data-provider/src/controllers/submissionController.ts
leoraba: Thanks for the suggestion. Code fixed.

joneubank (commented):
This all looks good, left a couple comments, if you can answer these we can approve this.

Would like to see some more tests on parts of this, in particular the merge code.

packages/data-provider/src/utils/submissionUtils.ts
```diff
@@ -304,8 +323,60 @@ const utils = (dependencies: BaseDependencies) => {
 			};
 		},
 
-		removeEntityFromSubmission: (submissionData: Record<string, SubmissionInsertData>, entityName: string) => {
-			return _.omit(submissionData, entityName);
+		removeItemsFromSubmission: (
+			submissionData: SubmissionData,
+			filter: { actionType: SubmissionActionType; entityName: string; index: number | null },
```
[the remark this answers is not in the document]
justincorrigible: https://github.com/overture-stack/lyric/pull/93#discussion_r1725699061

packages/data-model/docs/submissionCommit.md
```diff
@@ -0,0 +1,70 @@
...
+---
+sequenceDiagram
+    actor User
+
+    User->>+Lyric: /submission/category/:categoryId/commit/:submissionId
+    Lyric->>+DB: get Submission by submissionId
+    DB-->>-Lyric: Return Submission
+
+    Lyric->>Lyric: Check if submission exists
+    alt
```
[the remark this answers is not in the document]
leoraba: Code fixed. Added missing descriptions to `alt` command.

packages/data-model/docs/submissionCommit.md
```diff
@@ -0,0 +1,70 @@
...
+    
+        Lyric->>User: 400 Bad Request:<br />Submission '$submissionId}' not found
```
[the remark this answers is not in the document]
leoraba: good catch. Code fixed

packages/data-provider/src/repository/submittedRepository.ts
```diff
@@ -16,24 +16,26 @@ const repository = (dependencies: BaseDependencies) => {
...
-			systemId: data.systemId,
-			updatedAt: new Date(),
-			updatedBy: userName,
-		};
-		return await db.insert(auditSubmittedData).values(newAudit);
+	const auditDeleteSubmittedData = async (datas: SubmittedData[], userName: string) => {
+		datas.forEach(async (data) => {
+			const newAudit: NewAuditSubmittedData = {
+				action: AUDIT_ACTION.Values.DELETE,
+				comment: '',
```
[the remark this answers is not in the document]
leoraba: I don't know why `comment` is set empty. Anyways, I have a separate PR to restructure the Audit table where this code is changed.

packages/data-provider/src/utils/submissionUtils.ts
```diff
@@ -230,6 +232,23 @@ const utils = (dependencies: BaseDependencies) => {
...
+		/**
+		 * Merge two `Record<string, T[]>` objects into a single `Record<string, T[]>` object.
+		 * For each key in the records, the corresponding arrays from both records are concatenated.
+		 * @param record1 The first `Record<string, T[]>` object. If `undefined`, it is treated as an empty record.
+		 * @param record2 The second `Record<string, T[]>` object. If `undefined`, it is treated as an empty record.
+		 * @returns
+		 */
+		mergeRecords: <T>(
+			record1: Record<string, T[]> | undefined,
+			record2: Record<string, T[] | undefined>,
```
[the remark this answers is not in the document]
leoraba: not intentional. Code fixed

packages/data-provider/src/utils/submissionUtils.ts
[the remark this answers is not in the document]
leoraba: changed to `null` instead of `undefined` as we intentionally want to indicate that the variable `index` is set to "no value". Whereas setting to `undefined` could be interpreted the variable `index` has not been assigned a value yet.

leoraba:
PR updated. ready for code review

packages/data-provider/src/controllers/submissionController.ts
joneubank: a couple things here...

First, `req.query.index` could be 0, which would be falsey. This should explicitly check if index is undefined. `req.query.index !== undefined ? parseInt(req.query.index) : undefined;`

Second, is this a case where we need to differentiate between "no value provided" and "intentionally has no value". If we need to query the DB for records where the index === null (since DBs do use null values and can filter on those) then including `null` as an option is a good idea.  However, I think this is just a case where the `index` is an optional filter, which is best served by the type `index?: number` in a TS idiomatic way.

packages/data-provider/src/utils/submissionUtils.ts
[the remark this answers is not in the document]
joneubank: Responded to previous comment.

joneubank (approved):
Approved. I would still reconsider the use of `null`.

PR #95, opened 2024-08-21
Reactor Audit process to diff data before and after an update/delete

leoraba:
**Description of the changes:**
Track changes in Submitted Data using a more comprehensive format, with the ability to create a before and after data comparison.

- Created a new column on `audit_submitted_data` with the name of `dataDiff` to store the diff data in format:

```
{
    "old": {[k: string]: string | string[]},
    "new": {[k: string]: string | string[]}
}
```

- Additional changes on `audit_submitted_data` table to store the `submissionId` representing the change.
- All database changes are included on new Drizzle migration.
- These DB changes are intended to track changes on `Update` or `Delete` Submitted Data events, however currently only Delete Submitted Data endpoint is completed. Update Submitted Data endpoint is currently under development in a separate PR.

### DIFF
--- a/apps/server/swagger/schemas.yml
+++ b/apps/server/swagger/schemas.yml
@@ -101,6 +101,9 @@ components:
                   properties:
                     systemId:
                       type: string
+                    data:
+                      type: object
+                      description: Content to be deleted
         dictionary:
           type: object
           description: Dictionary used to validate the submission
@@ -307,37 +310,35 @@ components:
     ChangeHistoryRecord:
       type: object
       properties:
-        comment:
-          type: string
-          description: Comment of the event
         entityName:
           type: string
           description: Name of the Entity
         event:
           type: string
           description: Type of event
-        newData:
+          enum: ['UPDATE', 'DELETE']
+        dataDiff:
           type: object
-          description: New content of the record in JSON format
+          description: Captures the state of the data before the change as `old` and after the change as `new`
         newIsValid:
           type: boolean
-          description: New isValid value
-        oldData:
-          type: object
-          description: Old content of the record in JSON format
+          description: New  data isValid value
         oldIsValid:
           type: boolean
-          description: Old isValid value
+          description: Old data isValid value
         organization:
           type: string
           description: Name of the Organization
+        submissionId:
+          type: string
+          description: Submission ID the change belongs to
         systemId:
           type: string
-          description: System ID
-        updatedAt:
+          description: The unique identifier of the record changed
+        createdAt:
           type: string
           description: Date and time of creation
-        updatedBy:
+        createdBy:
           type: string
           description: User name who updated the submission
 
--- a/apps/server/swagger/schemas.yml
+++ b/apps/server/swagger/schemas.yml
@@ -400,6 +400,17 @@ components:
           type: string
           description: ID of the Submission
 
+    DeleteDataActiveSubmission:
+      type: object
+      properties:
+        records:
+          type: array
+          items:
+            $ref: '#/components/schemas/SubmittedDataRecord'
+        submissionId:
+          type: string
+          description: ID of the Submission
+
     GetSubmittedDataResult:
       type: object
       properties:
--- a/apps/server/swagger/schemas.yml
+++ b/apps/server/swagger/schemas.yml
@@ -63,20 +63,47 @@ components:
           description: ID of the Submission
         data:
           type: object
-          additionalProperties:
-            type: object
-            properties:
-              creator:
-                type: string
-                description: Name of user who created the submission
-              records:
+          properties:
+            inserts:
+              type: object
+              additionalProperties:
+                type: object
+                properties:
+                  creator:
+                    type: string
+                    description: Name of user who created the submission
+                  records:
+                    type: array
+                    items:
+                      type: object
+                      description: Content of the file in JSON format
+                  batchName:
+                    type: string
+                    description: Original filename  of the submission
+            updates:
+              type: object
+              additionalProperties:
                 type: array
                 items:
                   type: object
-                  description: Content of the file in JSON format
-              batchName:
-                type: string
-                description: Original filename  of the submission
+                  properties:
+                    systemId:
+                      type: string
+                    oldData:
+                      type: object
+                      description: Old Content
+                    newData:
+                      type: object
+                      description: New content
+            deletes:
+              type: object
+              additionalProperties:
+                type: array
+                items:
+                  type: object
+                  properties:
+                    systemId:
+                      type: string
         dictionary:
           type: object
           description: Dictionary used to validate the submission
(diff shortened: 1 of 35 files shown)


packages/data-model/docs/schema.dbml
leoraba, line 1: Changes on `audit` table to track data diff and submission_id responsible of the change

packages/data-model/migrations/meta/0007_snapshot.json
leoraba, line 1: files required by Drizzle to know when to generate migration files.

apps/server/swagger/schemas.yml
```diff
@@ -307,37 +310,35 @@ components:
...
           description: Name of the Entity
         event:
           type: string
           description: Type of event
-        newData:
+          enum: ['UPDATE', 'DELETE']
+        dataDiff:
           type: object
-          description: New content of the record in JSON format
+          description: Captures the state of the data before the change as `old` and after the change as `new`
```
justincorrigible, line 322: Thinking of Lyric as a "concurrent collaboration" service (multiple users could be updating a document at the same time), `Update` then should be seen as a way to input [operational transformations](https://en.wikipedia.org/wiki/Operational_transformation), meaning the records in Lyric are somewhat equivalent to stateful `deltas`.

This makes me want to advocate for using the word "previous" instead of "old".

(Leave it to me for all your semantic nitpicking needs 😆)

leoraba: I wasn't fully convinced to use `old/new` words, but now that you mention I like "previous" instead of "old", and perhaps "next" instead of "new"?

joneubank: "next" is ambiguous, as it seems to indicate a future state. previous/current makes gramatical sense, but when reading a trail of deltas "current" is also ambiguous. Perhaps previous/updated?

I was OK with old/new in a delta, as it simply describes the two states in the change.

leoraba: `previousState` and `newState` sounds better? or to make it short simply `previous` and `new`?

packages/data-provider/src/utils/submittedDataUtils.ts
```diff
@@ -16,15 +17,74 @@ import {
...
+				return diff;
+			}
+
+			if (!oldRecord) {
+				// oldRecord is null, all keys in newRecord are new
+				for (const key in newRecord) {
+					if (Object.prototype.hasOwnProperty.call(newRecord, key)) {
+						diff.new[key] = newRecord[key];
+					}
+				}
```
joneubank, line 52: I think I know why you included this if check, but for the record and to make sure this is needed...

Why are we checking that the record has the key, when you just got the key from the record? Is there some subset of properties that are included in the object iterator but are not included in `hasOwnProperty` for the record?

packages/data-provider/src/utils/submittedDataUtils.ts
```diff
@@ -16,15 +17,74 @@ import {
...
+				return diff;
+			}
+
+			// Both records are non-null, compare them
+			for (const key in oldRecord) {
+				if (Object.prototype.hasOwnProperty.call(oldRecord, key)) {
+					const oldValue = oldRecord[key];
+					const newValue = newRecord[key];
```
joneubank, line 70: No check was done to see if `newRecord` has this property... if it was a valid check on `oldRecord` then it is also valid on `newRecord`.

packages/data-provider/src/utils/submittedDataUtils.ts
```diff
@@ -16,15 +17,74 @@ import {
...
+
+					if (oldValue !== newValue) {
+						diff.old[key] = oldValue;
+						diff.new[key] = newValue ?? '';
+					}
```
joneubank, line 75: How do you differentiate between `newRecord[key]` being `''` and being removed (ie. `undefined`).

packages/data-provider/src/utils/submittedDataUtils.ts
```diff
@@ -16,15 +17,74 @@ import {
...
 	return {
+		/**
+		 * Compares two `DataRecord` objects and returns the differences between them.
+		 * @param oldRecord The original `DataRecord` object to compare.
+		 * @param newRecord The new `DataRecord` object to compare against the original.
+		 * @returns An object of type `DataDiff` containing the differences between `oldRecord` and `newRecord`.
+		 * 	The differing values are recorded with the `old` object containing the values
+		 * 	from `oldRecord` and the `new` object containing the corresponding values from `newRecord`.
+		 */
+		computeDataDiff: (oldRecord: DataRecord | null, newRecord: DataRecord | null): DataDiff => {
```
joneubank, line 38: We should write unit tests for this:
1. returns no change between 2 null values
2. returns no change between 2 objects with same values
3. reports new fields added to newRecord
4. reports fields removed from newRecord
5. reports values changed from oldRecord to newRecord
6. returns whole newRecord when compared to null oldRecord
7. returns whole oldRecord deleted when compared to null newRecord

joneubank (commented):
Basically ready to merge. Would like to see our diff method have tests. I put some comments on the naming discussion (`old` and `new` seem good to me), and I asked for clarification to one part of the diff code.

packages/data-provider/src/services/submissionService.ts
```diff
@@ -438,8 +441,14 @@ const service = (dependencies: BaseDependencies) => {
...
-		deletesSet.forEach((systemIdToDelete) => {
-			dataSubmittedRepo.deleteBySystemId(systemIdToDelete, userName);
+		// iterate if there are any record to be deleted
+		dataToValidate?.deletes?.forEach((item) => {
+			dataSubmittedRepo.deleteBySystemId({
+				submissionId: submission.id,
+				systemId: item.systemId,
+				diff: computeDataDiff(item.data as DataRecord, null),
+				userName,
+			});
```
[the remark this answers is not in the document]
leoraba: that's something really important that forgot to implement!. Looks like [Drizzle transactions](https://orm.drizzle.team/docs/transactions) can help, will work on this implementation

packages/data-provider/src/utils/submittedDataUtils.ts
leoraba: yes definitely need to add these tests. Thanks

packages/data-model/migrations/0007_create_audit_diff_column.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE "audit_submitted_data" RENAME COLUMN "updatedAt" TO "created_at";--> statement-breakpoint
+ALTER TABLE "audit_submitted_data" RENAME COLUMN "updatedBy" TO "created_by";--> statement-breakpoint
+ALTER TABLE "audit_submitted_data" ADD COLUMN "data_diff" jsonb;--> statement-breakpoint
```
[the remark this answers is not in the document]
leoraba: Yes it's defined on the DB model file `audit_submitted_data.ts`

packages/data-provider/src/utils/submittedDataUtils.ts
leoraba: This is a check to ensure that the key being iterated over in the `for...in` loop actually belongs to the newRecord object itself and is not an inherited property from its prototype like `toString` or others

packages/data-provider/src/utils/submittedDataUtils.ts
leoraba: if `newRecord` doesn't have this property we create the property as an empty string. in the following lines

packages/data-provider/src/utils/submittedDataUtils.ts
leoraba: Test cases were added to cover these and other scenearios

packages/data-provider/src/utils/dictionaryUtils.ts
leoraba, line 1: Moved most of this functions to `dictionaryService.ts` file. For the purpose of leaving on utils functions that doesn't require dependencies

packages/data-provider/src/utils/submittedDataUtils.ts
leoraba, line 1: This is a refactoring change. Moving all the functions out of the wrapper function, as most of them doesn't require any dependency. This will facilitate unit testing

leoraba:
PR updated to add some Unit Testing to submission/submittedData utils.

packages/data-provider/src/utils/submittedDataUtils.ts
joneubank: the removed case should be `undefined` instead of `''` no?

joneubank (approved):

packages/data-provider/src/utils/submittedDataUtils.ts
leoraba: Code changed to identify a removed field as `undefined`

PR #96, opened 2024-08-29
Edit submitted data

leoraba:
Task included on this PR:

- [x] :sparkles: New endpoint to Edit Submitted Data `PUT /submission/category/{categoryId}/data`
- [x] Update swagger description Edit Submitted Data
- [x] Accepts as input a `.tsv` file with the record and fields to be updated (it must include extra column called `systemId`)
- [x] Add `update` changes to an Active Submission
- [x] Cascade updates when changing ID fields.
- [x] Updates on ID fields deletes and creates a new record
- [x] `Commit` endpoint implements Edit records
- [x] Unit tests
- [x] Audit log events


apps/server/swagger/data-api.yml
```diff
@@ -252,3 +252,46 @@
         $ref: '#/components/responses/ServerError'
       503:
         $ref: '#/components/responses/ServiceUnavailableError'
+
+/data/category/{categoryId}/edit:
+  post:
+    summary: Edit Submitted Data. Returns an Active Submission with the records to be updated
+    tags:
+      - Admin
```
justincorrigible, line 260: curiosity: these tags, are they meant to imply "what user role level has access to it"? if so, what other meanings can one of these tags imply?

"admin" doesn't clearly convey what it means. a type of task, or who can do it, or what type of endpoint it is among other types?

apps/server/swagger/schemas.yml
```diff
@@ -122,10 +122,25 @@ components:
               type: string
         errors:
           type: object
-          additionalProperties:
-            type: array
-            items:
-              $ref: '#/components/schemas/ValidationError'
+          properties:
+            inserts:
```
justincorrigible, line 126: to clarify for my own understanding of what I'm looking at:
- is this a conjugated verb? (e.g. "the endpoint inserts X")
- or is it a plural? (e.g. "an array of something called inserts").

apps/server/swagger/data-api.yml
leoraba: The idea for this tag is to facilitate the visualization of a more "risky" group of operations that would require a higher privilege like an `admin` level. 
2 operations belong in this Admin group: **update** and **delete** Submitted data. Considering that these operations could break dependencies among already Submitted Data.

justincorrigible: understood! great explanation!
so this is a label to group "management" tasks, that are more destructive, and therefore ideally performed by an "admin"-level user or someone else who knows what they're doing, and which we're not authenticating in terms of level permissions.

in that order of ideas, I may recommend differentiating admin as a user group, from risky tasks that should be called something else to avoid confusion.

apps/server/swagger/schemas.yml
leoraba: This is the plural form of `insert`, consist on an key-value object with the new data to be inserted grouped by schema name. Similarly this applies for `update` and `delete`. As Submission can consist on insert, update or delete one or more records. 

e.g. 
```
 "data": {
    "inserts": {
      "sample": {
        "records": [
          {
            "study_id": "TEST_PRO",
            "sample_type": "Protein",
            "submitter_sample_id": "TM-01-1",
            "submitter_specimen_id": "LOD-04",
            "submitter_participant_id": "NR-30"
          }
        ],
        "batchName": "sample.tsv"
      }
    },
    "updates": {
      "participant": [
        {
          "new": {
            "date_of_birth": "1990-05"
          },
          "old": {
            "date_of_birth": "1970-05"
          },
          "systemId": "I7R9DH9OP59Y1EZE6PSN9"
        }
      ]
    },
    "deletes": {
      "specimen": [
        {
          "data": {
            "study_id": "TEST_PRO",
            "submitter_specimen_id": "LOD-01",
            "specimen_tissue_source": "Saliva",
            "specimen_collection_date": "2000-01",
            "submitter_participant_id": "NR-01",
            "specimen_anatomic_location": "C33.2"
          },
          "isValid": true,
          "systemId": "28UMRXZHF2XNQTIRQI4AZ",
          "entityName": "specimen",
          "organization": "oicr"
        }
      ]
    }
  },
```

errors are displayed the same way grouped by inserts, updates and deletes:
e.g. 
```
"errors": {
    "inserts": {
      "sample": [
        {
          "index": 0,
          "reason": "INVALID_BY_FOREIGNKEY",
          "fieldName": "submitter_participant_id",
          "fieldValue": "NR-30",
          "foreignSchema": {
            "fieldName": "submitter_participant_id",
            "schemaName": "participant"
          }
        }
      ]
    }
  }
```

apps/server/swagger/data-api.yml
leoraba: Just bringing more options to renaming this tag:
- "Data Management"
- "Data maintenance"
- "Data curation" 
- "Data control"
- "Critical Data Control"
Any other suggestion?

justincorrigible:
After reviewing/demoing the feature over zoom, we've identified some issues in the implementation of the endpoint, particularly in terms of idempotency of the operation, resulting in conflicting merges of the changesets.

@leoraba will address and update the PR soon

apps/server/swagger/data-api.yml
leoraba: Removed `admin` tag, as it was confusing. Left those affected endpoints under `Submission` tag as it's more representative

leoraba:
Renamed `Create`, `Edit` and `Delete` data endpoints to a more concise pattern. Keeping `/submission/category/{categoryId}` as the base of all three to maintain clear relationships between the submission process.
- Create Data (Submission Creation)
`POST /submission/category/{categoryId}/data`
- Edit Data (Submission Creation)
`PUT /submission/category/{categoryId}/data`
- Delete Data
`DELETE /submission/category/{categoryId}/data/{systemId}`

packages/data-provider/src/utils/submittedDataUtils.ts
```diff
@@ -209,3 +253,21 @@ export const transformmSubmittedDataToSubmissionDeleteData = (submittedData: Sub
 		return acc;
 	}, {});
 };
+
+export const updateSubmittedDataArray = (
+	submittedData: SubmittedData[],
+	editData: SubmissionUpdateData[],
+): SubmittedData[] => {
+	return submittedData.map((existingSubmittedData) => {
+		const found = editData.find((e) => e.systemId == existingSubmittedData.systemId);
```
joneubank, line 262: the rare `==` instead of `===`, is there a reason?

packages/data-provider/src/utils/types.ts
```diff
@@ -327,24 +335,34 @@ type Mutable<T> = {
 };
 
 export type MutableDataDiff = {
-	old: Mutable<DataRecord>;
-	new: Mutable<DataRecord>;
+	old: MutableDataRecord;
+	new: MutableDataRecord;
 };
 
+export type MutableDataRecord = Mutable<DataRecord>;
```
joneubank, line 342: type is used before being declared. TS doesn't care but it reads poorly. Don't worry about changing this, there is no need for Mutable type after the lectern client upgrade.

joneubank (commented):
Lots here, took a while to get through. No major concerns with the approach, just a couple code level recommendations for change.

packages/data-provider/src/utils/submissionUtils.ts
```diff
@@ -146,16 +159,178 @@ export const extractSchemaDataFromMergedDataRecords = (
...
+};
+
+/**
+ * Returns a filter to query the database used to find dependents records when the update record involves changes of an primary ID field
+ *
+ * @param schemaRelations An array of `SchemaChildNode` representing the schema relations for the entity. Each node contains information about parent-child relationships.
+ * @param updateRecord The update record containing old and new data. The function checks the `old` data to identify fields involved in the relationship.
+ * @returns
+ */
+export const getDependentsFilteronSubmissionUpdate = (
```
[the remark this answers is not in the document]
leoraba: I agree the name is not clear of what the function does. What about`filterRelationsForPrimaryIdUpdate`?

packages/data-provider/src/utils/submissionUtils.ts
```diff
@@ -146,16 +159,178 @@ export const extractSchemaDataFromMergedDataRecords = (
...
+ */
+export const filterRecordsByConflicts = <T, U>(
+	sourceData: Record<string, T[]>,
+	conflictData: Record<string, U[]>,
+	entitySelector: (item: T) => string,
+	conflictSelector: (item: U) => string,
+): Record<string, T[]> => {
+	if (!sourceData) {
+		return {};
+	}
```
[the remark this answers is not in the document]
leoraba: Removed this condition. Forgot to remove it after doing a refactoring

packages/data-provider/src/utils/submissionUtils.ts
```diff
@@ -146,16 +159,178 @@ export const extractSchemaDataFromMergedDataRecords = (
...
+ * This function can be used to either filter updates from deletes or deletes from updates, depending on the provided parameters.
+ * It removes records from the `sourceData` that have a matching `systemId` in the `conflictData`.
+ *
+ * @param sourceData - A record of the primary data (e.g., updates or deletes) to be filtered, grouped by entity name.
+ * @param conflictData - A record of data that might conflict (e.g., deletes or updates), grouped by entity name.
+ * @param entitySelector - A function to select the `systemId` from the source records.
+ * @param conflictSelector - A function to select the `systemId` from the conflict records.
+ * @returns A record of filtered source data, excluding records that conflict based on `systemId`.
+ */
+export const filterRecordsByConflicts = <T, U>(
```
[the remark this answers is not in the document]
leoraba: Renamed generics to `<SourceData, ConflictData>`

packages/data-provider/src/utils/submissionUtils.ts
```diff
@@ -126,13 +138,14 @@ export const checkFileNames = async (
...
  */
 export const determineIfIsSubmission = (
-	toBeDetermined: SubmittedDataReference | SubmissionReference,
-): toBeDetermined is SubmissionReference => {
-	return (toBeDetermined as SubmissionReference).type === MERGE_REFERENCE_TYPE.SUBMISSION;
+	toBeDetermined: SubmittedDataReference | NewSubmittedDataReference | EditSubmittedDataReference,
+): toBeDetermined is NewSubmittedDataReference | EditSubmittedDataReference => {
+	const type = (toBeDetermined as NewSubmittedDataReference | EditSubmittedDataReference).type;
+	return type === MERGE_REFERENCE_TYPE.NEW_SUBMITTED_DATA || type === MERGE_REFERENCE_TYPE.EDIT_SUBMITTED_DATA;
 };
```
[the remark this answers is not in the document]
leoraba: This is a great refactoring. Thanks for the suggestion!

packages/data-provider/src/utils/fileUtils.ts
```diff
@@ -78,3 +79,36 @@ export function getSizeInBytes(size: string | number): number {
 	// If value is a number it is assumed is in bytes.
 	return bytes.parse(size);
 }
+
+// sort files into validFiles and fileErrors based on correct file extension
+export async function processFiles(files: Express.Multer.File[]) {
+	const result = {
+		validFiles: [] as Express.Multer.File[],
+		fileErrors: [] as BatchError[],
+	};
```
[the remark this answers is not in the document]
leoraba: Added type as suggested!

packages/data-provider/src/utils/fileUtils.ts
```diff
@@ -78,3 +79,36 @@ export function getSizeInBytes(size: string | number): number {
 	// If value is a number it is assumed is in bytes.
 	return bytes.parse(size);
 }
+
+// sort files into validFiles and fileErrors based on correct file extension
+export async function processFiles(files: Express.Multer.File[]) {
```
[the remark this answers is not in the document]
leoraba: Documented this function using TSDoc

.vscode/settings.json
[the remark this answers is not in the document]
leoraba: Added `.vscode` to gitignore

packages/data-provider/src/utils/submittedDataUtils.ts
leoraba: Thanks for noticing. Changed to `===` as both should be same string type

packages/data-provider/src/utils/fileUtils.ts
```diff
@@ -78,3 +79,36 @@ export function getSizeInBytes(size: string | number): number {
...
+	const result = {
+		validFiles: [] as Express.Multer.File[],
+		fileErrors: [] as BatchError[],
+	};
+
+	for (const file of files) {
+		if (hasTsvExtension(file)) {
+			const fileHeaders = await readHeaders(file); // Wait for the async operation
```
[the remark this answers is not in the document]
leoraba: try catch added to handle errors reading files

packages/data-provider/src/services/submissionService.ts
```diff
@@ -101,33 +112,42 @@ const service = (dependencies: BaseDependencies) => {
...
+						data: record,
+						dictionaryCategoryId: categoryId,
+						entityName,
+						isValid: false, // By default, New Submitted Data is created as invalid until validation proves otherwise
+						organization: submission.organization,
+						originalSchemaId: submission.dictionaryId,
+						lastValidSchemaId: submission.dictionaryId,
+						systemId: generateIdentifier(entityName, record),
+						createdBy: userName,
+					}));
```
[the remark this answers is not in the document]
leoraba: The `commit` operation verifies that the Submission is valid.  However, instead of setting `lastValidSchemaId` before validating I've adjusted the code to set it after validation. That way when `lastValidSchemaId` is null and `isValid` is false means data has never been validated or never valid.  Otherwise `lastValidSchemaId` will be set when validation has passed.

leoraba:
PR updated and code refactored to solve comments.
PR ready for code review.

packages/data-provider/src/utils/types.ts
leoraba: This should be handled in next PR

packages/data-provider/src/utils/fileUtils.ts
```diff
@@ -16,7 +16,7 @@ export const ARRAY_DELIMITER_CHAR = '|';
  * @param {Express.Multer.File} file
  * @returns {boolean}
  */
-export const hasTsvExtension = (file: Express.Multer.File): boolean => !!file.originalname.match(/.*\.tsv$/);
+export const hasTsvExtension = (file: Express.Multer.File): boolean => !!file.originalname.match(/.*\.tsv$/i);
```
joneubank, line 19: nice change

joneubank (approved):
Great stuff. Thanks!

PR #112, opened 2024-11-21
Adding Endpoint for Record Validation

Azher2Ali:
This feature introduces a new endpoint for validating the existence of records within specific contexts, supporting applications that require robust data integrity. By ensuring records exist (or preventing duplicates), this endpoint enhances user experience and maintains data consistency.

**Endpoint Details:**
Method: GET
Path: `/validator/{categoryId}/{entityName}`

**Query Parameters:**

**studyId:** Identifier of the organization in Lyric, required as a string.
**value:** The value to validate, required as a string.
**categoryId:** Numeric ID of the category the record belongs to.
**entityName:** Name of the entity or analysis type (string) used in Song.

**Example Request:**
GET [http://localhost:3030/validator/{categoryId}/{entityName}?studyId={studyId}&value={value}](http://localhost:3030/validator/%7BcategoryId%7D/%7BentityName%7D?studyId=%7BstudyId%7D&value=%7Bvalue%7D)

**Expected Behavior:**
Record Exists: Returns 200 OK with an empty body.
Record Not Found: Returns 400 Bad Request with a response body:
```
{
   "error": "INVALID_VALUE",
   "message": "Description of why the value is invalid"
}
```
**Implemented Changes:**

- [x] Created a new ValidatorController to manage the validation endpoint.

- [x] Developed a ValidationService to handle logic for checking records within the submitted_data table.

- [x] Added a Router for routing the new endpoint.

- [x] Defined data models for ValidationRequest and ValidationResponse.

This endpoint’s design supports configuration-based category and entityName searches, offering flexibility for different validation contexts.

This feature is essential for applications that must verify records' existence within specific categories and organizations, providing a streamlined, reliable approach to validating data before further processing.

Also there are some pending changes related to Database Query, which will be added to this PR with our upcoming changes


packages/data-provider/src/services/validationService.ts
```diff
@@ -0,0 +1,49 @@
...
+            logger.debug(LOG_MODULE, 'Validating record', { categoryId, entityName, studyId, value });
+
+            try {
+                
+                const filterEntityNameSql = eq(submittedData.entityName, entityName);
+                const filter = eq(submittedData.systemId, studyId); //todo need to change
+
+                const resultCount = await db
+                    .select({ total: count() })
+                    .from(submittedData) //todo need to change
```
joneubank, line 30: Not sure what needs to be changed, can you elaborate? Should probably update the PR or leave a more detailed TODO comment (preferably do this work now ;) )

joneubank (changes requested):
Changes are clear and concise, nice work.

I would like to change two things:
1. let's have the response handled within the controller instead of throwing an error.
2. be more specific with the controller function name.

PR #113, opened 2024-11-22
feat: export migration scripts

leoraba:
## Description
Provide a function to run database migration.

## Package `Data Model` changes:
- The `migrate` function, located in `src/functions/migrate.ts`, is exported to enable running migrations from external applications.
- The script `scripts/migrate.js` is provided to execute migrations within the monorepo.
- The type `DbConfig` is being exported to define the database parameters.

### DIFF
--- a/packages/data-model/package.json
+++ b/packages/data-model/package.json
@@ -23,6 +23,10 @@
 	"files": [
 		"dist/"
 	],
+	"exports": {
+		".": "./index.js",
+		"./models": "./src/models/index.js"
+	},
 	"license": "AGPL-3.0-or-later",
 	"repository": {
 		"type": "git",
--- a/packages/data-model/package.json
+++ b/packages/data-model/package.json
@@ -9,8 +9,9 @@
 		"\n=====================  Build  =================": "",
 		"build:compile": "rimraf dist && tsc",
 		"build:migration": "drizzle-kit generate:pg --config=drizzle.config.ts",
+		"build:migration:copy": "copyfiles migrations/** migrations/**/**  dist/",
 		"build:dbml": "tsx scripts/dbmlGenerator.ts",
-		"build:all": "pnpm build:compile && pnpm build:migration && pnpm build:dbml",
+		"build:all": "pnpm build:compile && pnpm build:migration && pnpm build:migration:copy && pnpm build:dbml",
 		"\n==============  Dev environment  ==============": "",
 		"db:migrate:dev": "tsx scripts/migrate.ts",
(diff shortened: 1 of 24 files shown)


packages/data-model/package.json
```diff
@@ -9,8 +9,9 @@
 		"\n=====================  Build  =================": "",
 		"build:compile": "rimraf dist && tsc",
 		"build:migration": "drizzle-kit generate:pg --config=drizzle.config.ts",
+		"build:migration:copy": "copyfiles migrations/** migrations/**/**  dist/",
 		"build:dbml": "tsx scripts/dbmlGenerator.ts",
-		"build:all": "pnpm build:compile && pnpm build:migration && pnpm build:dbml",
+		"build:all": "pnpm build:compile && pnpm build:migration && pnpm build:migration:copy && pnpm build:dbml",
```
joneubank, line 14: 👍

packages/data-model/src/functions/migrate.ts
```diff
@@ -0,0 +1,36 @@
...
+const getRequiredConfig = (name: string) => {
+	const value = process.env[name];
+	if (!value) {
+		throw new Error(`No Environment Variable provided for required configuration parameter '${name}'`);
+	}
+	return value;
+};
+
+export async function migrate() {
+	console.log('Running your migrations...');
```
joneubank, line 18: Don't we have a logger in our dependencies?

packages/data-model/src/functions/migrate.ts
```diff
@@ -0,0 +1,36 @@
...
+
+	const sql = new pg.Pool({
+		host: getRequiredConfig('DB_HOST'),
+		database: getRequiredConfig('DB_NAME'),
+		password: getRequiredConfig('DB_PASSWORD'),
+		port: Number(getRequiredConfig('DB_PORT')),
+		user: getRequiredConfig('DB_USER'),
+	});
```
joneubank, line 26: Are we not exporting this in a way that uses dependencies like our config? When imported this puts requirements on the environment rather than just being a code library.

joneubank (commented):
Wondering if the exported migration code should take dependencies/config arguments.

packages/data-model/src/functions/migrate.ts
leoraba: You're right!. thanks for pointing this out. As this package `data-model` is imported in the `lyric-provider`, I will create here the `DbConfig` definition type to make it consistent along both packages.

leoraba:
PR updated. ready for code review

joneubank (approved):

Issue #120, opened 2024-12-04
New feature request: Retrieve previous Submissions by user


leoraba:
## Summary of request
Have the ability to retrieve previous submissions from current user.

## Desired solution
Be able to reuse one or all of the existing GET submission endpoints with additional/optional query parameter to return the current `Active` Submission or all previous ones.

## Details:
Having the existing GET submission Endpoints:
  1.  GET `/submission/{submissionId}`
  2.  GET `/submission/category/{categoryId}`
  3.  GET `/submission/category/{categoryId}/organization/{organization}`

### Endpoint 1
**Current behaviour:** Returns the Submission only if the status is `active`. 
**Expected:** Return Submission regardless of its status.  

### Endpoint 2: 
**Current behaviour:** Returns the Submission that is `active` in this category/organization.
**Expected:** 
  - Returns the Submission regardless of its status.
  - A new optional query param: `onlyActive` of type boolean. Default `false`. 
    - If no query parameter is sent, by default return the submission/submissions regardless of the submission `status`;
    - If query parameter is sent, return only the submission with `active` status.
  - Pagination functionality
  - Add `organization` as query parameter.

## Endpoint 3:
Move to `deprecated` endpoint, as new enhancements of endpoint #2 now fulfills this requirement

leoraba:
Add and optional query parameter to filter by user

leoraba:
This feature was completed and merge from PR [#121](https://github.com/overture-stack/lyric/pull/121)

PR #121, opened 2024-12-06
Feat #120 - Get previous submissions paginated

leoraba:
## Changes included:
- **Endpoint GET `submission/{submissionId}`:** 
  - returns the submission regardless of the submission status.
- **Endpoint GET `submission/category/{category}`:** 
  - returns all the submissions on a category regardless of the submission status. 
  - Added a new query parameter `onlyActive` (boolean) to return by active status. 
  - Added pagination parameters.
  - Added a new query parameter `organization` to filter by an specific organization name
- **Endpoint GET `submission/category/{category}/organization/{organization}`:** 
  - Marked as `deprecated` as it's replaced by submission by category endpoint.


joneubank:
I just read the PR description, but I assume these endpoints would be restricted to admins only? Maybe a user can use these but the results are restricted to their own submissions?

Perhaps this auth model hasn't been accounted for yet?

apps/server/swagger/parameters.yml
```diff
@@ -38,3 +38,17 @@ components:
           type: string
           enum: ['flat', 'compound']
         description: Optional query parameter to define the data format. Choose 'flat' for a simple, single-level collection of records, or 'compound' for a nested, schema-centric structure. The default value is 'flat'
+      OnlyActive:
+        description: Optional query parameter to filter results to include only active submissions. Default value is false
+        name: onlyActive
+        in: query
+        schema:
+          type: boolean
+        required: false
```
joneubank, line 47: Are there only two statuses? Would a filter by status be a better approach, allowing us to show closed/cancelled submissions or only succesfully submitted ones?

packages/data-provider/src/controllers/submissionController.ts
```diff
@@ -162,29 +164,51 @@ const controller = (dependencies: BaseDependencies) => {
...
+				const pageSize = parseInt(String(req.query.pageSize)) || defaultPageSize;
 
-				logger.info(LOG_MODULE, `Request Active Submission categoryId '${categoryId}'`);
+				logger.info(
+					LOG_MODULE,
+					`Request Submission categoryId '${categoryId}'`,
+					`pagination params: page '${page}' pageSize '${pageSize}'`,
+					`onlyActive '${onlyActive}'`,
+					`organization '${organization}'`,
+				);
```
joneubank, line 181: Do we have a request logging middleware? That middleware can capture this request information, it feels like a lot of extra work to log it the details of every request within the request.

packages/data-provider/src/repository/activeSubmissionRepository.ts
```diff
@@ -123,31 +134,79 @@ const repository = (dependencies: BaseDependencies) => {
...
+		 * Get Submissions by category
+		 * @param {number} categoryId - Category ID
+		 * @param {Object} paginationOptions - Pagination properties
+		 * @param {number} paginationOptions.page - Page number
+		 * @param {number} paginationOptions.pageSize - Items per page
+		 * @param {Object} filterOptions
+		 * @param {boolean} filterOptions.onlyActive - Filter by Active status
+		 * @param {string} filterOptions.userName - Filter by creator's username
+		 * @param {string} filterOptions.organization - Filter by Organization
 		 * @returns One or many Active Submissions
```
joneubank, line 146: This is a lot of parameter description... I've been skipping the parameter descriptions in cases like this where there isn't anything information to add beyond their name and type, since TS already provides name and type information. Additionally, I have found over time the TS declaration changes and the same details aren't updated 1-to-1 in the TS docs...

So my solution has been to always give a TS doc description but not to repeat all the function signature information, I only describe the params or return type if there is some interesting logic or information that the user of the function will want to know and that isn't captures in TS.

packages/data-provider/src/repository/activeSubmissionRepository.ts
```diff
@@ -123,31 +134,79 @@ const repository = (dependencies: BaseDependencies) => {
...
+					columns: getSubmissionColumns,
+					with: getSubmissionRelations,
+					orderBy: (submissions, { desc }) => desc(submissions.createdAt),
+					limit: pageSize,
+					offset: (page - 1) * pageSize,
 				});
 			} catch (error) {
-				logger.error(LOG_MODULE, `Failed querying Active Submission with relations`, error);
+				logger.error(LOG_MODULE, `Failed querying Submissions by category with relations`, error);
+				throw new ServiceUnavailable();
```
joneubank, line 174: Is this really service unavailable? I'm not sure what this error looks like, I would have expected to return something more generic like "Unexpected System Error" and the logged error explaining why the DB fetch failed.

packages/data-provider/test/utils/submissionUtils.spec.ts
joneubank, line 1: This file is very long, we should consider if there is a sensible way to reorganize this into smaller pieces that are easier to navigate.

joneubank (approved):

apps/server/swagger/parameters.yml
leoraba: This  boolean filter `onlyActive` facilitates to retrieve submission, when is `true` means submission that are `OPEN`, `VALID` or `INVALID`. 
When onlyActive is `false` means every submission including `CLOSED` and `COMMITED`

packages/data-provider/test/utils/submissionUtils.spec.ts
leoraba: this specific file is testing 19 functions related to submission utils. I've relocated 1 file per each function we're testing.

packages/data-provider/src/repository/activeSubmissionRepository.ts
leoraba: When an error occurs consuming an external service such as a DB or Lectern Server the error is caught and logged in the console, then and a generic `ServiceUnavailable` error is thrown:
```
{
  "error": "Service unavailable",
  "message": "Server is unable to access the necessary resources to process the request. Please try again later."
}
```
If a troubleshoot is required the details of the error would be visible only in the server terminal

packages/data-provider/src/repository/activeSubmissionRepository.ts
leoraba: I agree this tsdoc looks saturated, think this might be one of the largest. Although, I've tried to write Tsdocs to most of the functions following the same pattern like this.

packages/data-provider/src/controllers/submissionController.ts
leoraba: This is something I'll like to refactor in this repo. The issue is the instance of the logger Lyric is using is different to the logger the express server is using, causing inconsistent log formatting. I'd love to do this change in future PR to track isolated code changes.

leoraba:
Endpoints are not secured yet as Auth implementation is added in separate PR. As for now, these endpoints will only retrieve Submission of the user who's logged in, a filter to query by user can be added in a separate PR as a new feature.

PR #123, opened 2024-12-11
New Feature - Authentication Middleware Injection

leoraba:
## Enhancements and Features

- **Authentication Middleware Injection:** Added a new property in the provider package to support a custom auth implementation in the middleware. Applications using the Lyric provider can now pass a function  `customAuthHandler ` to enforce their authentication logic.

- **User Session utilization:** Updated Lyric to use the object `user` property from the incoming Request object to obtain user's name and permissions.
  - `username`:  used to track ownership of data submissions. user's name is recored in inserts, updates or deleletes.
  - `isAdmin`: used to allow write data privileges to any organization
  - `allowedWriteOrganizations`: used to allow write data privilege to an specific organization

- **Exported `UserSession` Type:** Introduced and exported the `UserSession` type to define user related properties. Lyric will use the `user` property from the incoming Request object to obtain the user name and for permissions

```
type UserSession = {
	username: string;
	isAdmin: boolean;
	allowedWriteOrganizations: string[];
}
```

- **Documentation Updates:** Updated the documentation to include guidelines on implementing and integrating a custom authentication middleware.


apps/server/src/middleware/auth.ts
```diff
@@ -0,0 +1,19 @@
...
+import { UserSession } from '@overture-stack/lyric';
+
+// Extend the Request interface to include a `user` property
+declare module 'express-serve-static-core' {
+	interface Request {
+		user?: UserSession;
+	}
+}
+
+export const authMiddleware = (req: Request, res: Response, next: NextFunction) => {
```
leoraba, line 12: a placeholder to implement auth middleware.

packages/data-provider/src/config/config.ts
```diff
@@ -48,6 +49,7 @@ export type AppConfig = {
 	limits: LimitsConfig;
 	logger: LoggerConfig;
 	schemaService: SchemaServiceConfig;
+	authMiddleware?: (req: Request, res: Response, next: NextFunction) => void;
```
leoraba, line 52: Applications that implements Lyric provider will pass a custom middleware for authentication

packages/data-provider/src/middleware/auth.ts
```diff
@@ -1,13 +1,49 @@
...
+ * If authentication is valid, it attaches the user information to the request object;
+ * Otherwise, it returns the appropriate error codes.
+ * @param authConfig
+ * @returns
  */
-export const auth = async (req: Request, res: Response, next: NextFunction) => {
-	// TODO: auth here
-	console.log(`auth`);
-	next();
+export const authMiddleware = (authConfig: AuthConfig) => {
```
leoraba, line 21: Middleware used when auth is enabled.
It uses the `customAuthHanlder` to performs the auth.
It returns to `next()` handler if user is valid authenticated or return a 401 or 403 error.

apps/server/src/auth/handler.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Function to implement authentication logic.
+ * This function is called by the authMiddleware to verify the user's authentication.
+ * It returns the authentication result, which includes the authentication status and user information.
+ *
+ * @param req - Express request object
+ * @returns User session result
+ */
+export const authHandler = (_req: Request) => {
```
leoraba, line 13: This is the implementation of an auth hanlder used to provider Guest access.

packages/data-provider/src/middleware/auth.ts
```diff
@@ -1,13 +1,49 @@
...
+		try {
+			const authResult: UserSessionResult =
+				typeof authConfig.customAuthHandler === 'function'
+					? authConfig.customAuthHandler(req)
+					: { authStatus: 'invalid-auth' };
+
+			switch (authResult.authStatus) {
+				case 'authenticated':
+					req.user = authResult.user;
+					return next();
```
leoraba, line 35: if user is authenticated, set user details into request

README.md
```diff
@@ -88,6 +88,7 @@ The Environment Variables used for this application are listed in the table bell
...
 | `PORT`                      | Server Port.                                                                                                                                                                                                                                                                                         | 3030                                   |
 | `UPLOAD_LIMIT`              | Limit upload file size in string or number. <br>Supported units and abbreviations are as follows and are case-insensitive: <br> - b for bytes<br> - kb for kilobytes<br>- mb for megabytes<br>- gb for gigabytes<br>- tb for terabytes<br>- pb for petabytes<br>Any other text is considered as byte | '10mb'                                 |
+| `ALLOWED_ORIGINS`           | Specifies a list of permitted origins for Cross-Origin Resource Sharing (CORS). These origins, separated by commas, are allowed to make requests to the server, ensuring only trusted domains can access resources.                                                                                  |                                        |
```
demariadaniel, line 91: Example value could be helpful here

apps/server/src/auth/handler.ts
demariadaniel: I don't know the whole project context, so I'm wondering why the readme example for authHandler contains actual auth logic, but the actual `authHandler` implementation just returns a hardcoded value

packages/data-provider/README.md
```diff
@@ -8,56 +8,146 @@
...
+		url: 'https://api.lectern-service.com', // URL of the schema service
 	},
 };
 
 
 const lyricProvider = provider(appConfig);
 ```
 
-Use any of the resources available on provider on a Express server:
+### Auth Custom Hanlder
```
demariadaniel, line 56: Typo, should be `Handler`

packages/data-provider/README.md
```diff
@@ -8,56 +8,146 @@
...
 ## Support & Contributions
 
 - Developer documentation [docs](https://github.com/overture-stack/lyric/blob/main/packages/data-provider/docs/add-new-resources.md)
 - Filing an [issue](https://github.com/overture-stack/lyric/issues)
 - Connect with us on [Slack](http://slack.overture.bio)
 - Add or Upvote a [feature request](https://github.com/overture-stack/lyric/issues/new?assignees=&labels=&projects=&template=Feature_Requests.md)
+
+```
+
+```
```
demariadaniel, line 153: This code block looks unused?

demariadaniel:
A couple recommended tidying tasks but all in all this looks like a very clean implementation. I'll leave some time for team members closer to the project to take a look, but it looks good to me!

apps/server/src/auth/handler.ts
leoraba: The handler implemented in this `apps/server` package is a basic implementation for demonstration, however, the details how to customize it is described in [readme](https://github.com/overture-stack/lyric/pull/123/files#diff-e466f736b4bcb12a226f20fb6d671f6a26057fea5b635cddbcef98c8dddcf468R58) file

apps/server/src/server.ts
```diff
@@ -1,23 +1,40 @@
...
 
 app.use(helmet());
 
+app.use(
+	cors({
+		origin: function (origin, callback) {
+			if (!origin) {
+				// allow requests with no origin
+				// (like mobile apps or curl requests)
+				return callback(null, true);
```
justincorrigible, line 28: do we want to add a feature flag (env var config) to enable this, and have it disabled by default?
doesn't feel right to just have it "open" like this

apps/server/src/server.ts
```diff
@@ -35,6 +52,6 @@ app.use('/health', healthRouter);
 
 app.use(errorHandler);
 // running the server
-app.listen(serverConfig.port, () => {
-	console.log(`Starting Express server on http://localhost:${serverConfig.port}`);
+app.listen(port, () => {
+	console.log(`Starting Express server on http://localhost:${port}`);
```
justincorrigible, line 56: the server doesn't know its host, and "localhost" may be misleading.
```suggestion
	console.log(`Starting ExpressJS server on port ${port}`);
```

packages/data-provider/src/middleware/auth.ts
```diff
@@ -1,13 +1,62 @@
 import { NextFunction, Request, Response } from 'express';
 
+export type AuthStatus = 'authenticated' | 'no-auth' | 'invalid-auth';
```
justincorrigible, line 3: is "no auth" meant as "auth is disabled for the server" or "auth is enabled and user hasn't logged in"? think we may need more semantic nuance here

packages/data-provider/README.md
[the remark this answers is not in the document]
leoraba: I think providing our database migration script allows integrators to ensure Lyric will keep track of codebase and database schema in sync. For example, integrators over time can jump to latest version of Lyric, and they will be sure the database schema will be updated as well.

packages/data-provider/src/middleware/auth.ts
leoraba: The relevant property from `user` to Lyric is an `username` (described in `UserSession` [type](https://github.com/overture-stack/lyric/pull/123/files#diff-a3e4b8f8555de3488f9106e93c2482f1ccbf7e835dc3654306cc4277e80d62e5R5-R7))  Lyric uses this value to set ownership to submissions and audit records. For that reason, this could be any user identifier (could be an id, name, or email, etc.) in any case the implementer can set any value to this property.

justincorrigible: thanks for the additional details! not for this PR, but instincts tell me that it would be useful for integrators being able customize this part

packages/data-provider/src/middleware/auth.ts
leoraba: What I mean with `no-auth` is user didn't provide any auth on the request. Although I admit this can lead to a confusion. 
I propose to change this variable to a more meaning full variable. Instead, a variable named like `errorCode`  where implementers can set what error is sent 401, 403, etc.

justincorrigible: yeah, that sounds right. Error server codes are already standardized, and we can well take advantage of that

apps/server/src/server.ts
leoraba: Thanks for noticing this. In regards security this should be disabled by default. 
And could be converted in to a variable named `CORS_ENABLED` where when `true` only registered domains can access the server, otherwise rejected (default)

packages/data-provider/src/middleware/auth.ts
leoraba: code changed to allow custom `errorCode` and `errorMessage`

packages/data-provider/src/middleware/auth.ts
```diff
@@ -1,13 +1,53 @@
...
+ * @returns
  */
-export const auth = async (req: Request, res: Response, next: NextFunction) => {
-	// TODO: auth here
-	console.log(`auth`);
-	next();
+export const authMiddleware = (authConfig: AuthConfig) => {
+	return (req: Request, res: Response, next: NextFunction) => {
+		// proceed to the next middleware or route handler if auth is disabled
+		if (!authConfig.enabled) return next();
```
justincorrigible, line 36: for consistency, `if` statements should all have curlies
```suggestion
		if (!authConfig.enabled) {
			return next();
		}
```

apps/server/src/auth/handler.ts
justincorrigible: perhaps It'd be useful then to rename these as defaultAuthHandler and defaultAuthResult? (side note: instinctively wanted to say "response" instead of "result", but realized that may be context-dependent)

packages/data-provider/README.md
```diff
@@ -14,34 +14,111 @@ npm i @overture-stack/lyric
...
+    try {
+		// Verify the token using a public key
+		const publicKey = process.env.JWT_PUBLIC_KEY!;
+		const decodedToken = jwt.verify(token, publicKey);
+
+		return {
+			user: {
+				username: decodedToken.username,
+				isAdmin: decodedToken.isAdmin,
+				allowedWriteOrganizations: decodedToken.scopes,
```
justincorrigible, line 96: are all the scopes "write"? no "read"-only available?

packages/data-provider/README.md
```diff
@@ -14,34 +14,111 @@ npm i @overture-stack/lyric
...
+	}
+};
+```
+
+Add the handler to your `AppConfig` object.
+
+```javascript
+import { AppConfig, provider, UserSession } from '@overture-stack/lyric';
+
+const appConfig: AppConfig = {
```
justincorrigible, line 113: same suggestion as earlier, making this `defaultAuthConfig` would help readability.

also enabled:true by default doesn't feel undebatable. what's the rationale?

justincorrigible: gah, we're still in the readme. never mind

packages/data-provider/README.md
justincorrigible: even if it's just a readme's "example", this seems confusing

leoraba: I've updated this README to clarify these properties. in essence, when `user` object is provided the user has READ access to every endpoint., and only restricts write access to the organizations listed on `allowedWriteOrganizations `, or write access to all organization when `isAdmin` is `true`

apps/server/src/auth/handler.ts
leoraba: I've removed the auth handler on `apps/server` the reason it is not really implementing any auth logic, however it's described in README how implementers should use auth handlers.

packages/data-provider/test/utils/auth.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { expect } from 'chai';
+import { describe, it } from 'mocha';
```
justincorrigible, line 2: TODO: switch to Node test runner, to decrease external dependencies
e.g. `node:test` and `node:assert`

justincorrigible (approved):

Issue #110, opened 2024-11-15
Configuration of external validation endpoint


leoraba:
## Summary of request
Validation endpoint requires a basic configuration in order to perform checks of ID fields. 

## Details
* Add required configuration properties to `AppConfig` and `BaseDependencies` . For instance `validator` should contain an array of **category, entity and field** to enable search. 
* Every request into the validator endpoint should check if category, entity and field exists in configuration, If not, return 404 Not found response.

Issue #109, opened 2024-11-15
External validation endpoint


leoraba:
## Summary of request
This feature introduces an endpoint designed to verify the existence of a record with a specified value. The endpoint will be capable of performing checks within a given category, organization, and entity to ensure the existence of a record.

This feature is essential for applications that require validation of records within specific contexts, enhancing data integrity and user experience by preventing duplicate entries or ensuring the presence of required records.

## Functionality Details:

* The endpoint is a GET method, it will receive query parameters that specify the `studyId` and `value` to be checked, as well as the `categoryId` and `entityName` in which the record should be searched. Example:
```
GET http://localhost:3030/validator/{categoryId}/{entityName}?studyId={studyId}&value{value}
```
Where:
    - `{categoryId}` is the ID of the category the record belongs. Number type.
    - `{entityName}` is a string, also known as Analysis Type in Song
    - `{studyId}` is a string, also known as Organization in Lyric
    - `{value}` is a string, value to check

* If the record is found, the endpoint will respond with a 200 OK status with empty body, confirming the existence of the record.
* If the record does not exist, the endpoint will return a 400 Bad Request status, indicating that the specified value was not found. E.g.
```
{
   "error": "INVALID_VALUE",
   "message": "Any string message describing why the value is invalid"
}
```
* Endpoint should require to be configured to enable searching by category and entityName

PR #124, opened 2024-12-12
Add Validation Endpoint with Configurable Category, Entity, and Field Validation

Azher2Ali:
# Description
This feature introduces a configurable endpoint to validate the existence of a record with a specified value. It supports checks within a specific category, organization, and entity to ensure data integrity by preventing duplicates and confirming the presence of required records.

### Key Features:

Validation Endpoint:
Path: `/validator/category/{categoryId}/entity/{entityName}/exists?organization={organization}&value{value}`

**Method: GET**
**Behavior:**

If the record exists in the submitted_data table, returns **200 OK** with body: 
```
{
  "message": "Record found"
}
```

If the record does not exist, returns **404 Not found** with an error response:

```
{
   "error": "Not found",
   "message": "The specified value was not found"
}
```

### Configuration Requirements:

- Added configuration properties in AppConfig to define the valid combinations of category, entity, and field.
-  Requests to the endpoint validate that category, entity, exist in the configuration. If not, returns a **400 Bad request** with error response:
```
{
   "error": "Bad Request",
   "message": "Validation is not enabled for categoryId '${categoryId}' and entityName '${entityName}'"
}
```

### Implemented Changes:
- [x]  Added router  `ValidatorRouter`  for routing the new endpoint.
- [x] Created a `ValidatorController` to handle the new endpoint.
- [x]  Developed a `ValidationService` to check records in the submitted_data table.
- [x]  Defined data models for ValidationRequest and ValidationResponse.

### Configuration Validation Behavior
- Ensures the categoryId, entityName, and fieldName combinations are present in the configuration.
- Returns 400 Not Found if validation fails.
- This feature is crucial for applications needing configurable record validation, offering flexibility while ensuring strict data integrity standards.

### Related issues:
- issue #109 
- issue #110


packages/data-provider/src/core/provider.ts
```diff
@@ -50,13 +52,15 @@ const provider = (configData: AppConfig) => {
...
 			submittedData: submittedDataRouter(baseDeps),
+			validator: validationRouter(baseDeps),
 		},
 		controllers: {
 			audit: auditController(baseDeps),
 			category: categoryController(baseDeps),
 			dictionary: dictionaryController(baseDeps),
 			submission: submissionController(baseDeps),
 			submittedData: submittedDataController(baseDeps),
+			validator: validationService(baseDeps)
```
leoraba, line 63: `controller` should have a `validationController` and `services` a `validationService`

apps/server/src/server.ts
```diff
@@ -27,6 +27,7 @@ app.use('/category', lyricProvider.routers.category);
 app.use('/data', lyricProvider.routers.submittedData);
 app.use('/dictionary', lyricProvider.routers.dictionary);
 app.use('/submission', lyricProvider.routers.submission);
+app.use('', lyricProvider.routers.validator);
```
leoraba, line 30: this empty route looks weird, can you have here `/validator` and remove it from the router?

packages/data-provider/src/utils/schemas.ts
```diff
@@ -387,3 +387,57 @@ export const dataGetBySystemIdRequestSchema: RequestValidation<
...
+	categoryId: categoryIdSchema,
+	entityName: entityNameSchema,
+});
+
+export interface validationPathParams extends ParamsDictionary {
+	categoryId: string;
+	entityName: string;
+}
+
+export const validationRequestSchema1: RequestValidation<
```
leoraba, line 429: Looks like there are 2 similar validation Request Schemas `validationRequestSchema1` and `validationRequestSchema` which one are we using?

packages/data-provider/src/services/validationService.ts
```diff
@@ -0,0 +1,45 @@
...
+
+                const resultCount = await db
+                    .select({ total: count() })
+                    .from(submittedData) 
+                    .where(
+                        and(
+                            eq(submittedData.dictionaryCategoryId, categoryId),
+                            eq(submittedData.entityName, entityName), 
+                            eq(submittedData.organization, studyId),
+                            sql`${submittedData.data} @> ${JSON.stringify({ [entityName]: value })}::jsonb`
```
leoraba, line 33: In this query, `[entityName]` should come from a configurable variable, this variable should be a string type that matches a field name within the entity of the category.

For instance, 
the categoryId `1`, has entity `participant`,  within participant properties there is one called `participant_id` field.

Via Configuration we set category: `1` entity: `participant` and field: `participant_id`

So if Request is category `1`. and entity `participant`, we will know based on configuration given to use field `participant_id` in this query. The resulting query will be:
`WHERE categoryId=1 AND entityName='particpant' AND data @> '{"particpant_id": "123"}'`

leoraba (commented):
Left some comments but more importantly about the configuration part. Also more details on issue #110

joneubank (commented):
Looks generally good to me now that all required files are here. I'll leave it to @leoraba to give final approval on this.

packages/data-provider/src/services/validationService.ts
```diff
@@ -0,0 +1,53 @@
...
+					parseSQON({
+						op: 'in',
+						content: { fieldName: field, value: [value] },
+					}),
+				);
+
+				const totalRecords = await getTotalRecordsByCategoryIdAndOrganization(categoryId, organization, {
+					sql: filterSqon,
+					entityNames: [entityName],
+				});
```
joneubank, line 42: This works fine, but is some very awkward syntax to work with as a developer... does the SQON builder client not work for this?

Something like: 

```ts
const filterSqon = SQONBuilder.in(field, value);
```

joneubank (approved):
Looks good, accomplishes our initial design goal. I think we need to update this design slightly for how we configure the end point, but that will require an additional task. This is good to be merged.

packages/data-provider/src/services/validationService.ts
leoraba: This could also be separated in 2 parts for better understanding:
```
const sqonFilter = SQONBuilder.default.in(field, value);
const sqlFilter = convertSqonToQuery(sqonFilter);
```
I've changed the code to address this.

packages/data-provider/src/utils/validator.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export const findValidatorEntry = ({
+	validatorConfig,
+	categoryId,
+	entityName,
+}: {
+	validatorConfig: ValidatorConfig;
+	categoryId: string;
+	entityName: string;
+}) => {
```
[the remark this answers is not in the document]
leoraba: Added return type `ValidatorEntry | undefined` as suggested.

README.md
```diff
@@ -56,23 +56,24 @@ A model-agnostic, tabular data submission system designed to manage and validate
...
+| `PORT`                      | Server Port                                                                                                                                                                                                                                                                                                                                                                                                 | 3030                                   |
+| `VALIDATOR_CONFIG`          | Defines a configurable endpoint for validating the existence of a record with a specific value. The configuration is provided as an array of objects, where each object must include 'categoryId', 'entityName', and 'fieldName'. These entries determine which entity-field combinations are enabled for validation. (Example: `[{"categoryId": "1", "entityName": "sample", "fieldName": "sample_id" }]`) |                                        |
```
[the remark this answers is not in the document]
leoraba: created ticket #150 to detail the work refactoring this configuration.

PR #146, opened 2025-06-10
#26: Fix data loss on Submission with multiple files

leoraba:
## Summary

Fix issue where processing a submission with multiple files falls in a race condition where some data get lost.

## Issues
- [#26](https://github.com/Pan-Canadian-Genome-Library/clinical-submission/issues/26)

### Data Provider Package
The Submission Service now handles batch processing of multiple files or entity datasets, updating submission data accordingly where each submission is updated in the database only once, and the entire dataset is validated a single time per submission. 
Making this process in a batch based approach ensures consistency and helps to prevent race conditions.

#### Extras included:
- Added Unit tests to new functions, to make sure response is as expected.
- Splitting large files by relocating functions in separate files.

## Readiness Checklist

- [X] **Self Review**
  - I have performed a self review of code
  - I have run the application locally and manually tested the feature
  - I have checked all updates to correct typos and misspellings
- [X] **Formatting**
  - Code follows the project style guide
  - Autmated code formatters (ie. Prettier) have been run
- [X] **Local Testing**
  - Successfully built all packages locally
  - Successfully ran all test suites, all unit and integration tests pass
- [X] **Updated Tests**
  - Unit and integration tests have been added that describe the bug that was fixed or the features that were added
- [X] Documentation
  - All new environment variables added to `.env.schema` file and documented in the README
  - All changes to server HTTP endpoints have open-api documentation
  - All new functions exported from their module have TSDoc comment documentation

PR #148, opened 2025-06-24
#26: Fix Submit new data - Data loss on Submission with multiple files

leoraba:
## Summary

Fix issue where processing a submission with multiple files falls in a race condition where some data get lost.

> **Note:** This PR is part of the PR [#146](https://github.com/overture-stack/lyric/pull/146).  In order to split larger PRs and make it easier for code review, this PR only include fixes for the Insert New Data process.

## Issues
- [#26](https://github.com/Pan-Canadian-Genome-Library/clinical-submission/issues/26)

### Data Provider Package
The Submission Service now handles batch processing of multiple files or entity datasets, updating submission data accordingly where each submission is updated in the database only once, and the entire dataset is validated a single time per submission. 
Making this process in a batch based approach ensures consistency and helps to prevent race conditions.

#### Extras included:
- Added Unit tests to new functions, to make sure response is as expected.

## Readiness Checklist

- [X] **Self Review**
- [X] **Formatting**
  - Autmated code formatters (ie. Prettier) have been run
- [X] **Local Testing**
- [X] **Updated Tests**
- [X] Documentation

### DIFF
--- a/packages/data-provider/src/utils/formatUtils.ts
+++ b/packages/data-provider/src/utils/formatUtils.ts
@@ -115,10 +115,6 @@ export const deepCompare = (obj1: unknown, obj2: unknown): boolean => {
 	return true;
 };
 
-export const isNotNull = <T>(value: T): value is Exclude<T, null> => {
-	return value !== null;
-};
-
 // Helper function to check if an object is a plain object
 function isObject(obj: unknown): obj is Record<string, unknown> {
 	return typeof obj === 'object' && obj !== null && !Array.isArray(obj);
--- a/packages/data-provider/src/utils/recordsParser.ts
+++ b/packages/data-provider/src/utils/recordsParser.ts
@@ -2,7 +2,7 @@ import { type DataRecord, parse, type Schema } from '@overture-stack/lectern-cli
 import type { SubmissionInsertData } from '@overture-stack/lyric-data-model/models';
 
 import { getSchemaByName } from './dictionaryUtils.js';
-import { convertRecordToString, isNotNull, notEmpty } from './formatUtils.js';
+import { convertRecordToString, notEmpty } from './formatUtils.js';
 import { createBatchResponse } from './submissionResponseParser.js';
 import type { EntityData, SchemasDictionary } from './types.js';
 
@@ -55,6 +55,6 @@ export const parseRecordsToInsert = (
 
 				return [schemaName, createBatchResponse(schemaName, parsedRecords)];
 			})
-			.filter(isNotNull),
+			.filter(notEmpty),
 	);
 };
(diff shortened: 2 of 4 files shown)


packages/data-provider/src/controllers/submissionController.ts
```diff
@@ -312,8 +312,7 @@ const controller = ({
 				const username = user?.username || '';
 
 				const resultSubmission = await service.submit({
-					records: payload,
-					entityName,
+					data: { [entityName]: payload },
```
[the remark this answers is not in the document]
leoraba: This is significant change, where `submit` function can handle multiple entities at the same time, allowing the payload to be treated as a batch and being validated only once, avoiding raicing condition.

packages/data-provider/src/utils/formatUtils.ts
```diff
@@ -115,6 +115,10 @@ export const deepCompare = (obj1: unknown, obj2: unknown): boolean => {
 	return true;
 };
 
+export const isNotNull = <T>(value: T): value is Exclude<T, null> => {
+	return value !== null;
+};
+
```
joneubank, line 121: This function hardly seems necessary... but we can add it since a lot of the methods here are also not needed but seem convenient.

In the newer versions of TS you don't need to declare `value is ...` in the type signature if the function returns a boolean after narrowing the type. The TS compiler will infer the type narrowing correctly. This is the preferred way to write these function signatures since developers can write the type narrowing wrong, or the function can be modified later causing the type narrowing to be invalid. Best to let the compiler infer it. Generally, consider `value is X` similar to asserting a type `as X`.

packages/data-provider/test/utils/submission/parseRecordsToInsert.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+			{
+				name: 'product',
+				fields: [
+					{ name: 'id', valueType: 'number' },
+					{ name: 'name', valueType: 'string' },
+				],
+			},
+		],
+	};
+	it('should build insert records correctly for valid input', () => {
```
joneubank, line 32: This statement is unclear. If this failed I would not know what the test is attempting to do and likely wouldn't be able to fix it.

packages/data-provider/src/utils/recordsParser.ts
```diff
@@ -0,0 +1,60 @@
...
+};
+
+/**
+ * Parses raw records into typed data records based on the provided schema.
+ * @param dataRecords An array of unprocessed records with unknown value types.
+ * @param schema The schema definition used to convert and validate each record's fields.
+ * @returns An array of valid typed data records.
+ */
+export const convertToTypedRecords = (dataRecords: Record<string, unknown>[], schema: Schema): DataRecord[] => {
+	return Object.values(dataRecords).map(convertRecordToString).map(getSchemaParser(schema)).filter(notEmpty);
```
joneubank, line 27: 👍 for `.filter(notEmpty)`

joneubank (approved):
I would consider improving that one test description that I left a comment on. The rest of this change is simple and effective.

PR #166, opened 2025-11-04
69 - Able to update an entity with wrong fields from another entity

leoraba:
# Description
This PR fixes and issue where creating a submission to edit data by systemId on the wrong entity does not return any error.

## Code changes:
- The Submission is now marked as `INVALID` when `systemId` does not exists in the right entity.
- The submission includes in its response the systemId with empty values `{}`  (means no data were found):
  **Example:**
```json
"data": {
    "deletes": {},
    "inserts": {},
    "updates": {
      "diagnosis": [
        {
          "new": {},
          "old": {},
          "systemId": "1065Y2R2GWUHKD1U47E3V"
        },
        {
          "new": {},
          "old": {},
          "systemId": "DOESNOTEXIST"
        }
      ]
    }
  },
```
- The submission includes in its response a list of errors to identify the records that are invalid:
  **Example:**
```json
{
  "errors": {
    "updates": {
      "diagnosis": [
        {
          "index": 0,
          "reason": "UNRECOGNIZED_VALUE",
          "fieldName": "systemId",
          "fieldValue": "1065Y2R2GWUHKD1U47E3V"
        },
        {
          "index": 1,
          "reason": "UNRECOGNIZED_VALUE",
          "fieldName": "systemId",
          "fieldValue": "DOESNOTEXIST"
        }
      ]
    }
  }
}
```
- Unit tests updated to include new functions added

## Issue related:
- https://github.com/Pan-Canadian-Genome-Library/clinical-submission/issues/69

### DIFF
--- a/packages/data-provider/src/utils/submissionUtils.ts
+++ b/packages/data-provider/src/utils/submissionUtils.ts
@@ -245,45 +245,47 @@ export const groupSchemaErrorsByEntity = (input: {
 }): SubmissionErrors => {
 	const { resultValidation, dataValidated } = input;
 
-	const submissionSchemaErrors: SubmissionErrors = {};
 	if (resultValidation.valid) {
 		return {};
 	}
+
+	const submissionSchemaErrors: SubmissionErrors = {};
 	resultValidation.details.forEach((dictionaryValidationError) => {
 		const entityName = dictionaryValidationError.schemaName;
-		if (dictionaryValidationError.reason === 'INVALID_RECORDS') {
-			const validationErrors = dictionaryValidationError.invalidRecords;
-
-			const hasErrorByIndex = groupErrorsByIndex(validationErrors);
-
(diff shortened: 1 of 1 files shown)


packages/data-provider/src/services/submission/processor.ts
```diff
@@ -421,6 +423,39 @@ const processor = (dependencies: BaseDependencies) => {
...
+				if (!submissionSchemaErrors.updates) {
+					submissionSchemaErrors.updates = {};
+				}
+
+				if (!submissionSchemaErrors.updates[entityName]) {
+					submissionSchemaErrors.updates[entityName] = [];
+				}
+
+				submissionSchemaErrors.updates[entityName].push(
+					createUnrecognizedFieldBatchError({
```
joneubank: Is this the correct error name? The field `systemId` is recognized, but the value is not found for this entity. This is an invalid value, not unrecognized field.

PR #182, opened 2026-01-09
Dictionary migration - Part 1 - Database changes

### LINKED ISSUE #175: [Part 1 - Data migration] Trigger Data category migration.
# Description
Based on main ticket [169](https://github.com/overture-stack/lyric/issues/169) complete the following tasks:

## Main tasks
PR 1: https://github.com/overture-stack/lyric/pull/182
- Apply the database changes described on main ticket. update Drizzle models and migrations.
- Create a `dictionaryMigrationRepository` with functions: `save`, `update`, `getMigrationById`, `getMigrationsByCategoryId` (with filters and pagination)

PR 2: https://github.com/overture-stack/lyric/pull/183
- Update response of `POST /dictionary/register` endpoint accordingly to include `migrationId` when applicable.
- Implementation:
  - create a `migrationService` with functions: `getActiveMigrationByCategoryId`, `initiateMigration` (new migration or add retry), `finalizeMigration` (failed or completed), 
  - trigger the data main migration process async (`performCommitSubmissionAsync`), more details provided in main ticket.
  - log data errors in audit table.
  - when migration is complete update status to COMPLETED, and submission as COMMITTED
- block commit operation when migration is running

leoraba:
# Summary
This PR is part 1 of a series of changes for the Dictionary Migration epic https://github.com/overture-stack/lyric/issues/175.

Includes database changes only.


packages/data-provider/src/repository/dictionaryMigrationRepository.ts
```diff
@@ -0,0 +1,98 @@
...
+		 * Returns the inserted record's ID
+		 **/
+		save: async (data: NewDictionaryMigration): Promise<number> => {
+			try {
+				const savedMigration = await db
+					.insert(dictionaryMigration)
+					.values(data)
+					.returning({ id: dictionaryMigration.id });
+				logger.debug(LOG_MODULE, `Dictionary Migration for categoryId '${data.categoryId}' saved successfully`);
+				return savedMigration[0].id;
```
joneubank: Unsafe array access, `savedMigration[0]` could be undefined.

packages/data-provider/src/repository/dictionaryMigrationRepository.ts
```diff
@@ -0,0 +1,98 @@
...
+		/** Update an existing Dictionary Migration in Database */
+		update: async (migrationId: number, data: Partial<DictionaryMigration>): Promise<number> => {
+			try {
+				const updatedMigration = await db
+					.update(dictionaryMigration)
+					.set(data)
+					.where(eq(dictionaryMigration.id, migrationId))
+					.returning({ id: dictionaryMigration.id });
+				logger.debug(LOG_MODULE, `Dictionary Migration with id '${migrationId}' updated successfully`);
+				return updatedMigration[0].id;
```
joneubank: Unsafe array access, `updatedMigration[0]` could be undefined.

joneubank: maybe we would consider enabling the tsconfig property for noUncheckedIndexAccess (not sure if this would cause many errors in the existing code or not).

packages/data-provider/src/utils/types.ts
```diff
@@ -34,10 +34,13 @@ export const SUBMISSION_STATUS = {
 } as const;
 export type SubmissionStatus = ObjectValues<typeof SUBMISSION_STATUS>;
 
+export const MIGRATION_STATUS = z.enum(['IN-PROGRESS', 'COMPLETED', 'FAILED']);
```
joneubank: Should we use underscores instead of hyphens for UPPER_CASE enums? This seems to be the standard for all caps constant values, ie: `IN_PROGRESS`

joneubank (changes requested):
Good approach, minor change suggestions.

packages/data-provider/src/repository/dictionaryMigrationRepository.ts
leoraba: this is good catch, adding this check in tsconfig detected many errors in a total of 15 files, but it's worth adding this check.

packages/data-provider/src/repository/dictionaryMigrationRepository.ts
leoraba: check added

leoraba:
PR updated to address feedback with unchecked index and migration enum status

JamesTLopez (commented):
I have a small thought but optional, otherwise lgtm

packages/data-provider/src/repository/submittedRepository.ts
```diff
@@ -351,7 +357,7 @@ const repository = (dependencies: BaseDependencies) => {
 							filterEntityNameSql,
 						),
 					);
-				return resultCount[0].total;
+				return resultCount[0]?.total ?? 0;
```
JamesTLopez: Would there be an advantage to deconstructing `resultCount` and checking if its `undefined` to throw an error instead of the nullish coalesing. Undefined isn't really 0 records, in this case drizzle failed to fetch for us and failed to throw an error.

 I don't see why drizzle would return an `undefined` from this:

```ts
const [resultCount] = await db
  .select({ total: count() })
  .from(submittedData)
  .where(
  	and(
  		eq(submittedData.dictionaryCategoryId, categoryId),
  		eq(submittedData.organization, organization),
  		filter?.sql,
  		filterEntityNameSql,
  	),
  );
  
  if (!resultCount) {
  throw new Error('Somehow drizzle returned undefined but didnt throw an error');
  }
  
  return resultCount.total;

```

Current approach works, just a thought

joneubank: Wouldn't a fail to fetch throw an error? The drizzle behaviour is meant to return undefined when nothing found (0 results), but through an error if the fetch fails. As I understand it (please correct if I am wrong), this change isn't needed.

joneubank (approved):
Thanks for the updates!

packages/data-provider/src/repository/submittedRepository.ts
leoraba: the `undefined` response from drizzle in this query is very uncommon to happen, because if any error happens during drizzle fetch it just throws an error immediately. 

However in case this happens it wouldn't hurt add an extra validation as suggested. will change the code for this query.

PR #183, opened 2026-01-09
Dictionary migration - Part 2 - Execution migration

### LINKED ISSUE #175: [Part 1 - Data migration] Trigger Data category migration.

leoraba:
# Summary
Dictionary migration implementation, **PR 2** of https://github.com/overture-stack/lyric/issues/175 

### This PR depends on:
- https://github.com/overture-stack/lyric/pull/182


apps/server/swagger/schemas.yml
```diff
@@ -471,6 +471,10 @@ components:
         version:
           type: string
           description: Version of the Dictionary
+        migrationId:
+          type: string
+          description: ID of the Migration if applicable
+          allowEmptyValue: true
```
leoraba, line 477: New optional field in the response registering a new dictionary.

packages/data-provider/src/repository/submittedRepository.ts
```diff
@@ -392,28 +398,38 @@ const repository = (dependencies: BaseDependencies) => {
...
-						oldIsValid: boolean;
-						submissionId: number;
+						data: Partial<SubmittedData>;
+						audit: {
+							dataDiff: DataDiff;
+							errors?: DictionaryValidationRecordErrorDetails[];
+							isMigration: boolean;
+							oldIsValid: boolean;
+							submissionId: number;
+						};
```
leoraba, line 427: Adding 2 new properties `errors` and `isMigration` in the args, grouping values used for auditing

packages/data-provider/src/workers/workerPoolManager.ts
```diff
@@ -84,6 +84,17 @@ export const createWorkerPool = (configData: AppConfig): WorkerFunctions => {
 				// This ensures the main thread is not affected by worker errors and can continue processing other tasks.
 			}
 		},
+		dictionaryMigration: async (input) => {
```
leoraba, line 87: this is the entry point function to be used to start a working thread to handle the dictionary migration.

packages/data-provider/src/services/migrationService.ts
```diff
@@ -0,0 +1,217 @@
...
+
+	/**
+	 * **This function is designed to be executed in a worker thread.**
+	 * Performs the Submitted data validation for a given migration,
+	 * it iterates over all organizations and validates the submitted data for each of them.
+	 * @param migrationId The ID of the migration to perform
+	 * @param userName The name of the user that initiated the migration (for audit purposes)
+	 * @returns void
+	 */
+	const performMigrationValidation = async ({
```
leoraba, line 160: added a tsdoc to indicate the reason why this function is intended to by called on a worker thread

packages/data-provider/src/services/submission/submissionProcessor.ts
```diff
@@ -326,114 +327,125 @@ const createSubmissionProcessor = (dependencies: BaseDependencies) => {
...
+					if (record.id) {
+						const oldIsValid = record.isValid;
 						const inputUpdate: Partial<SubmittedData> = {};
-						const submisionUpdateData = dataToValidate.updates && dataToValidate.updates[data.systemId];
+
+						const submisionUpdateData = dataToValidate.updates?.[record.systemId];
 						if (submisionUpdateData) {
-							logger.info(LOG_MODULE, `Updating submittedData system ID '${data.systemId}' in entity '${entityName}'`);
-							inputUpdate.data = data.data;
+							logger.info(
```
joneubank: Should probably be debug level, we expect a lot of records to be updated in a migration.

packages/data-provider/src/services/submission/submissionProcessor.ts
joneubank, line 1: I was having a hard time reviewing the many changes to `performCommitSubmissionAsync`, the submission process is long with many steps and since its all inline with many checks for empty values or errors there is a lot of error handling mixed in with actual actions. Its likely that the processes could be separated into functions for each step, or more comments added to outline the process that is performed. This is likely a refactoring process separate from this PR.

packages/data-provider/src/workers/dictionaryMigrationWorker.ts
```diff
@@ -0,0 +1,24 @@
...
+			userName,
+		});
+	} catch (error) {
+		console.error('Error performing migration validation:', error);
+		migrationService.finalizeMigration({
+			migrationId,
+			status: 'FAILED',
+			userName,
+		});
+	}
```
joneubank, line 24: This finalize will likely fail for the same reason that caused it to throw an error the first time, but there is no error handling in place for this call. Will this crash the server? What is the expected outcome here?

packages/data-provider/src/services/migrationService.ts
```diff
@@ -0,0 +1,217 @@
...
+	 */
+	const finalizeMigration = async ({
+		migrationId,
+		status,
+		userName,
+	}: {
+		migrationId: number;
+		status: Extract<MigrationStatus, 'COMPLETED' | 'FAILED'>;
+		userName: string;
+	}): Promise<number> => {
```
joneubank: Why return the `migrationId`?, thats one of the input parameters. Also, the only place that calls this does not use the returned value.

This could easily return `Promise<void>`.

Should this return a Result instead of throwing?

packages/data-provider/src/services/dictionaryService.ts
```diff
@@ -141,12 +144,19 @@ const dictionaryService = (dependencies: BaseDependencies) => {
 				updatedBy: username,
 			});
 
+			const migrationId = await initiateMigration({
+				categoryId: updatedCategory.id,
+				fromDictionaryId: foundCategory.activeDictionaryId,
+				toDictionaryId: savedDictionary.id,
+				userName: username || '',
+			});
```
joneubank: initiateMigration throws on failure. Should this be in a try/catch? Maybe it should return a Result?

JamesTLopez (commented):
<s> I have a small thought but optional, otherwise lgtm </s>

joneubank (changes requested):
Generally looks good, have a couple small items to look at. Thanks!

packages/data-provider/src/services/submission/submissionProcessor.ts
leoraba: changed to `debug` level, we don't need to log every record, there's a `info` log already as a summary of the total # of records changed.

packages/data-provider/src/services/migrationService.ts
leoraba: this function and others in this files, have been refactored to return Result instead, to avoid throwing errors.

packages/data-provider/src/services/dictionaryService.ts
leoraba: `initiateMigration` function now returns a `Result` object, so now the error is returned as a string value.

packages/data-provider/src/workers/dictionaryMigrationWorker.ts
leoraba: Code is refactored so each function on the migrationService returns a Result object, error is handled as string

packages/data-provider/src/services/submission/submissionProcessor.ts
leoraba: fo facilitate the code refactoring, i'll create a new PR to split this functions and add unit tests.

leoraba:
PR has ben updated based on feedback. Ready for code review.

joneubank (approved):

Issue #169, opened 2025-11-13
Request - Data category migration to new dictionary


joneubank:
## Summary of request
We need a mechanism that will update the dictionary of a data category from one version of a dictionary to a new version. During this migration, we need to revalidate all records vs the new dictionary, and when errors are found we need to record in a separate table a list of migration errors that exist. This will allow us to show all records in their stored and unmodified state including which dictionary version they are valid for, but also look at each record showing what issues they have versus the latest dictionary.

PR #200, opened 2026-04-16
Dictionary Migration - Get migration endpoints

### LINKED ISSUE #176: [Part 2 - Data migration] Get Migrations endpoints
# Description 
Based on main ticket https://github.com/overture-stack/lyric/issues/169 complete the following tasks:

## Tasks
- Create 2 new endpoints to view general info of migrations.
- Create 1 new endpoint to view records updated by migration.

leoraba:
# Description
This PR creates 3 new GET endpoints to retrieve migrations information.
- GET /migration/category/{categoryId}:
- GET /migration/{migrationId}
- GET /migration/{migrationId}/records

## Related tickets:
- https://github.com/overture-stack/lyric/issues/176
- https://github.com/overture-stack/lyric/issues/169

## This PR depends on:
- https://github.com/overture-stack/lyric/pull/182
- https://github.com/overture-stack/lyric/pull/183


apps/server/swagger/schemas.yml
```diff
@@ -453,6 +438,88 @@ components:
...
+        submissionId:
+          type: number
+          description: ID of the Submission used for the migration
+        status:
+          type: string
+          description: Status of the Migration
+          enum: ['IN-PROGRESS', 'COMPLETED', 'FAILED']
+        retries:
+          type: number
+          description: Number of retries of the migration
```
joneubank: Do we explain "retries" anywhere? this is a confusing data point.

packages/data-provider/src/controllers/migrationController.ts
```diff
@@ -0,0 +1,106 @@
...
+	migrationDataRequestSchema,
+	migrationsByCategoryIdRequestSchema,
+} from '../utils/schemas.js';
+
+const controller = (dependencies: BaseDependencies) => {
+	const migrationService = createMigrationService(dependencies);
+	const { logger } = dependencies;
+	const LOG_MODULE = 'MIGRATION_CONTROLLER';
+	const defaultPage = 1;
+	const defaultPageSize = 20;
```
joneubank: Maybe should be global constants defined with the pagination types?

packages/data-provider/src/controllers/migrationController.ts
```diff
@@ -0,0 +1,106 @@
...
+
+	return {
+		getMigrationById: validateRequest(migrationByIdRequestSchema, async (req, res, next) => {
+			try {
+				const migrationId = Number(req.params.migrationId);
+
+				logger.info(LOG_MODULE, `Request Migration id '${migrationId}'`);
```
joneubank, line 25: should probably be debug level

packages/data-provider/src/controllers/migrationController.ts
```diff
@@ -0,0 +1,106 @@
...
+
+				const submission = await migrationService.getMigrationById(migrationId);
+
+				if (!submission) {
+					throw new NotFound('Submission not found');
```
joneubank: should we be logging that the migration was not found?

packages/data-provider/src/controllers/migrationController.ts
joneubank: This comment applies this to all similar request detail logging in this controller.

packages/data-provider/src/controllers/migrationController.ts
```diff
@@ -0,0 +1,106 @@
...
+				const page = parseInt(String(req.query.page)) || defaultPage;
+				const pageSize = parseInt(String(req.query.pageSize)) || defaultPageSize;
+
+				logger.info(LOG_MODULE, `Request Migrations by category Id '${categoryId}'`);
+
+				const submissions = await migrationService.getMigrationsByCategoryId(categoryId, { page, pageSize });
+
+				if (submissions.result.length === 0) {
+					throw new NotFound('Submissions not found');
+				}
```
joneubank: If the category ID exists, this should return an empty list.

If the category ID does not exist, then it should return NotFound.

packages/data-provider/src/repository/auditRepository.ts
```diff
@@ -30,20 +30,38 @@ const repository = (dependencies: BaseDependencies) => {
...
+			filterArray.push(eq(auditSubmittedData.organization, organization));
+		}
+
+		if (submissionId) {
+			filterArray.push(eq(auditSubmittedData.submissionId, submissionId));
+		}
+
+		if (newIsValid !== undefined) {
+			filterArray.push(eq(auditSubmittedData.newDataIsValid, newIsValid));
+		}
```
joneubank, line 63: Do we have auditRepository integration tests that confirm the filters are behaving as expected?

packages/data-provider/src/repository/dictionaryMigrationRepository.ts
```diff
@@ -1,19 +1,73 @@
...
+	const migrationWithRelationsColumns = {
+		category: {
+			columns: {
+				id: true,
+				name: true,
+			},
+		},
+		fromDictionary: dictionarySummaryColumns,
+		toDictionary: dictionarySummaryColumns,
+	} as const satisfies Record<string, { columns: BooleanTrueObject }>;
```
joneubank: Any reason why the second const `migrationWithRelationsColumns` doesn't test against the same type? Comparing `BooleanTrueObject` to `Partial<Record<keyof Dictionary, boolean>>`

packages/data-provider/src/repository/dictionaryMigrationRepository.ts
```diff
@@ -75,29 +131,102 @@ const repository = (dependencies: BaseDependencies) => {
...
-		): Promise<DictionaryMigration[]> => {
+		): Promise<{
+			result: MigrationRecordWithRelations[];
+			metadata: { totalRecords: number; errorMessage?: string };
+		}> => {
 			const { page, pageSize } = paginationOptions;
 
 			const { status, fromDictionaryId, toDictionaryId } = filterOptions;
 			try {
+				const whereConditions: SQL | undefined = and(
```
joneubank: Should not declare type, it is inferred from the called function.

packages/data-provider/src/repository/dictionaryMigrationRepository.ts
```diff
@@ -75,29 +131,102 @@ const repository = (dependencies: BaseDependencies) => {
...
+
+				if (!migration) {
+					logger.debug(LOG_MODULE, `No migration found with id '${migrationId}' when fetching migration records`);
+					return {
+						result: [],
+						metadata: { totalRecords: 0 },
+					};
+				}
+
+				const newIsValid = options.isInvalid !== undefined ? !options.isInvalid : undefined;
```
joneubank, line 201: Should the function argument just be changed to `isValid`?

packages/data-provider/src/routers/migrationRouter.ts
```diff
@@ -0,0 +1,29 @@
...
+	router.use(urlencoded({ extended: false }));
+	router.use(json());
+
+	router.use(authMiddleware(authConfig));
+
+	router.get('/:migrationId', migrationController(baseDependencies).getMigrationById);
+
+	router.get('/category/:categoryId', migrationController(baseDependencies).getMigrationsByCategoryId);
+
+	router.get('/:migrationId/records', migrationController(baseDependencies).getMigrationRecords);
```
joneubank: Just define the migration controller once and then reference it:

```ts
import createMigrationController from '../controllers/migrationController.js';

const router = ({...}) => {

  const migrationController = createMigrationController(baseDependencies);

  router.get('/:migrationId', migrationController.getMigrationById);
  router.get('/category/:categoryId', migrationController.getMigrationsByCategoryId);
  router.get('/:migrationId/records', migrationController.getMigrationRecords);
```

packages/data-provider/src/services/migrationService.ts
```diff
@@ -47,16 +55,17 @@ const migrationService = (dependencies: BaseDependencies) => {
 	 * @param categoryId
 	 * @returns
 	 */
-	const getActiveMigrationByCategoryId = async (categoryId: number): Promise<DictionaryMigration | null> => {
+	const getActiveMigrationByCategoryId = async (categoryId: number): Promise<MigrationRecordWithRelations | null> => {
```
joneubank, line 67: Needs to explicitly state why this would return null. (I assume: when there is no active migration for this category).

This function also propagates errors from the underlying services, so it should probably mention that it could fail with errors thrown.

packages/data-provider/src/services/migrationService.ts
```diff
@@ -67,6 +76,79 @@ const migrationService = (dependencies: BaseDependencies) => {
 		}
 	};
 
+	const getMigrationById = async (migrationId: number): Promise<MigrationRecordWithRelations | undefined> => {
```
joneubank: Similar comment as previous... this needs to explain why it sometimes returns `undefined` instead of the returned type.

While I prefer using `undefined` for missing values, the rest of the Service functions return `null` instead. It is confusing to have both values used in an ad-hoc way. Let's consistently use either `null` or `undefined`

packages/data-provider/src/services/migrationService.ts
```diff
@@ -67,6 +76,79 @@ const migrationService = (dependencies: BaseDependencies) => {
...
+		} catch (error) {
+			logger.error(LOG_MODULE, `Error retrieving migration with id '${migrationId}'`, error);
+			throw error;
+		}
+	};
+
+	const getMigrationsByCategoryId = async (
+		categoryId: number,
+		paginationOptions: PaginationOptions,
+	): Promise<{ metadata: { totalRecords: number; errorMessage?: string }; result: MigrationRecordWithRelations[] }> => {
```
joneubank: Might be useful to create a type alias for this return and export it for use where this function is consumed.

packages/data-provider/src/services/migrationService.ts
```diff
@@ -67,6 +76,79 @@ const migrationService = (dependencies: BaseDependencies) => {
...
+			return {
+				metadata: {
+					totalRecords: migrations.metadata.totalRecords,
+				},
+				result: migrations.result.map(formatMigrationSummary),
+			};
+		} catch (error) {
+			logger.error(LOG_MODULE, `Error retrieving migrations for categoryId '${categoryId}'`, error);
+			throw error;
+		}
```
joneubank: This can be a troubling pattern to fall into, where every function that could throw an error just logs the error then throws it. You can get a long log where every function in the call stack logs the same error with no new information added. Generally, we want to log errors where we catch and handle them.

We could instead make a design decision here, or in any function within this call stack, that the caught error can be handled here instead of thrown again. We can replace our return value either with void return type if there is exactly only one expected failure condition, or replace the return type with a Result that has a list of known failure conditions.

packages/data-provider/src/services/migrationService.ts
```diff
@@ -67,6 +76,79 @@ const migrationService = (dependencies: BaseDependencies) => {
...
+	const getMigrationRecords = async (
+		migrationId: number,
+		options: {
+			page: number;
+			pageSize: number;
+			entityNames?: string | string[];
+			organizations?: string | string[];
+			isInvalid?: boolean;
+		},
+	): Promise<{ result: AuditDataResponse[]; metadata: { totalRecords: number; errorMessage?: string } }> => {
```
joneubank: Another place for a helpful type alias.

packages/data-provider/src/utils/migrationResponseFormatter.ts
```diff
@@ -0,0 +1,25 @@
...
+export type MigrationSummary = MigrationRecordWithRelations & {
+	invalidRecords?: number;
+};
+
+/**
+ * Order the properties of the Migration Summary Record
+ * @param migration
+ * @returns
+ */
+export const formatMigrationSummary = (migration: MigrationSummary): MigrationSummary => ({
```
joneubank, line 13: I assume this is to make the return type from the route nicely formatted for presentation. If this is important to enforce, there should be a unit test for this on the route.

Don't spend much time on this next comment, but: this may be a good use for [branded types](https://www.learningtypescript.com/articles/branded-types). You can create a Branded type `MigrationSummaryFormatted` and have the controller require that it returns this branded type, that will enforce that this function is called before returning. This will enforce the rule through the type system, but since its only used for the route return type there's not a lot gained for the amount of effort here, so its not worth doin. Writing a test is the best way to enforce this.

joneubank (changes requested):
Minor changes here. Good work.

apps/server/swagger/schemas.yml
leoraba: I've updated the description for this field `"Number of times the migration has run. When a migration fails, it can be retried by registering again the same dictionary."`

However, a proper documentation is added in the following PR, where the "retry" functionality is implemented

packages/data-provider/src/controllers/migrationController.ts
leoraba: moved this constants to a new shared file.

packages/data-provider/src/controllers/migrationController.ts
leoraba: we add a `info` log for every request in every controller because we're not using express request logger.

packages/data-provider/src/controllers/migrationController.ts
leoraba: 👍 code changed as suggested

packages/data-provider/src/controllers/migrationController.ts
leoraba: code changed to return NotFound when category does not exists, also changed to return same error in other endpoints of this controller

packages/data-provider/src/repository/dictionaryMigrationRepository.ts
leoraba: created a `WithColumns<T>` type to be used along the code

packages/data-provider/src/repository/dictionaryMigrationRepository.ts
leoraba: type removed

packages/data-provider/src/repository/dictionaryMigrationRepository.ts
leoraba: in the `Audit` table each registry has `newIsValid` and `oldIsValid` columns to identify if the record has changed its validation, for that reason `new` is used for this constant

packages/data-provider/src/services/migrationService.ts
leoraba: Updated tsdocs to clarify return values and thrown errors

packages/data-provider/src/services/migrationService.ts
leoraba: for consistency, will return null when migration is not found, and will update tsdocs

packages/data-provider/src/services/migrationService.ts
leoraba: added a type `PaginatedResult<T>` to be used in this and other functions

packages/data-provider/test/unit/utils/migrationResponseFormatter.spec.ts
leoraba, line 1: unit tests added for migration formatter functions

packages/data-provider/src/utils/migrationResponseFormatter.ts
leoraba: for the purpose of this PR, I added unit test for this function to verify it, I would considerate using branded types as suggested in future PRs

packages/data-provider/src/services/migrationService.ts
leoraba: Removed the try-catch here as the error is unhandled in this function, it should be handled by the upper function (controller) that uses this function, repeated this logic in other functions in this Service.

packages/data-provider/src/repository/auditRepository.ts
leoraba: The integration test is broken due to a previous change in the worker pool, it requires additional work out of the scope of this PR and should be fixed in a separate PR and extended to cover this function

leoraba:
PR has been updated based on the feedback received, it is now ready for code review.

packages/data-provider/src/controllers/migrationController.ts
```diff
@@ -0,0 +1,117 @@
...
+					pagination: {
+						currentPage: page,
+						pageSize: pageSize,
+						totalPages: Math.ceil(migrationsResult.metadata.totalRecords / pageSize),
+						totalRecords: migrationsResult.metadata.totalRecords,
+					},
+					records: migrationsResult.result.map(formatMigrationSummary),
+				};
+
+				return res.send(response);
```
JamesTLopez: We should send a status 200

packages/data-provider/src/repository/auditRepository.ts
```diff
@@ -69,98 +87,84 @@ const repository = (dependencies: BaseDependencies) => {
...
-		 * @param {string} organization
-		 * @param {object} filterOptions
-		 * @param {string} filterOptions.entityName
-		 * @param {string} filterOptions.eventType
-		 * @param {string} filterOptions.startDate
-		 * @param {string} filterOptions.endDate
-		 * @param {string} filterOptions.systemId
+		 * and additional filters
+		 * @param {number} categoryId Category ID to filter the Audit records
+		 * @param {object} filterOptions Additional filters and pagination options
```
JamesTLopez: Instead of {object} `AuditFilterOptions` might be better for the tsDocs

packages/data-provider/src/services/migrationService.ts
```diff
@@ -30,7 +37,7 @@ const migrationService = (dependencies: BaseDependencies) => {
 		migrationId: number;
 		status: Extract<MigrationStatus, 'COMPLETED' | 'FAILED'>;
 		userName: string;
-	}): Promise<Result<null, string>> => {
+	}): AsyncResult<null, string> => {
```
JamesTLopez, line 40: Should we use the `AsyncResult` pattern for the remainder of the services in this file? Or is this only used for create/update actions?

packages/data-provider/src/utils/migrationResponseFormatter.ts
```diff
@@ -0,0 +1,38 @@
+import type { MigrationRecordWithRelations } from '../repository/dictionaryMigrationRepository.js';
+import type { AuditRepositoryRecord, MigrationAuditRecord } from './types.js';
+
+export type MigrationSummary = MigrationRecordWithRelations & {
+	invalidRecords?: number;
+};
+
+/**
+ * Order the properties of the Migration Summary Record
+ * @param migration
```
JamesTLopez:
```suggestion
 * @param {MigrationSummary} migration - The migration summary to format.
```

apps/server/swagger/schemas.yml
```diff
@@ -62,20 +62,23 @@ components:
...
+      type: object
+      properties:
+        currentPage:
+          type: number
+        pageSize:
+          type: number
+        totalPages:
+          type: number
+        totalRecords:
+          type: number
```
joneubank, line 75: :+1:

apps/server/swagger/schemas.yml
```diff
@@ -453,6 +443,121 @@ components:
...
+              type: string
+            version:
+              type: string
+        submissionId:
+          type: number
+          description: ID of the Submission used for the migration
+        status:
+          type: string
+          description: Status of the Migration
+          enum: ['IN-PROGRESS', 'COMPLETED', 'FAILED']
```
joneubank, line 538: `status` reports that the migration completed... is there anythign that indicates if any records have errors after the migration?

packages/data-provider/src/controllers/migrationController.ts
joneubank: Really minor point, but we should be using `res.json(response)` instead of `.send`.

They will be almost identical, or even exactly identical, in most cases. The point though is to indicate that we INTEND to be responding with JSON. If our response is not an object (like a string) then `.send(someString)` will happily send a response that does not have Content-Type application/json, and TS will not prevent us from returning this value.

There's also some weird cases where if you do `send(400)` it will set the 400 as the http status and send an empty body. Its best to avoid the "smart" logic in `.send` and use `.json` when you intend to return json.

packages/data-provider/src/repository/auditRepository.ts
joneubank: Responding to:
> The integration test is broken due to a previous change in the worker pool, it requires additional work out of the scope of this PR and should be fixed in a separate PR and extended to cover this function

Can you create an issue and link it?

packages/data-provider/src/repository/auditRepository.ts
joneubank: Responding to James' comment:
> Instead of {object} AuditFilterOptions might be better for the tsDocs

Alternately, TS captures our function interfaces. It is entirely acceptable to remove the param annotations, or just the type information, from the TSDoc. These properties are useful if there is something about the parameter that needs to be described - like maybe it is used in a confusing way by the function logic and the caller needs to know that.

packages/data-provider/src/repository/dictionaryMigrationRepository.ts
joneubank: I forgot this. Thanks for clarification.

packages/data-provider/src/services/migrationService.ts
joneubank: Long term, I would like to. Not required for scope for this PR.

joneubank (approved):
This looks very good. Thanks!

packages/data-provider/src/repository/auditRepository.ts
leoraba: Issue [#203](https://github.com/overture-stack/lyric/issues/203) created.

A Draft [PR #202](https://github.com/overture-stack/lyric/pull/202) has already fixed the broken integration tests and currently in progress to add Dictionary Migration tests

packages/data-provider/src/controllers/migrationController.ts
leoraba: added `.status(200)` in the response

leoraba: code change to respond with `.json(response)` as suggested.

packages/data-provider/src/repository/auditRepository.ts
leoraba: I just removed the type in the tsdocs to avoid unnecessary information in the docs

packages/data-provider/src/services/migrationService.ts
leoraba: for simplicity just using the `AsyncResult ` pattern in the actionables functions: initiate, perform and finalize migrations.

packages/data-provider/src/utils/migrationResponseFormatter.ts
leoraba: docs updated as suggested

apps/server/swagger/schemas.yml
leoraba: there is `invalidRecords` I missed updating the openapi docs, now it reflects the right response

PR #201, opened 2026-05-04
retry dictionary registration

### LINKED ISSUE #177: [Part 3 - Data migration] Retry migration
# Description 
Based on main ticket https://github.com/overture-stack/lyric/issues/169 complete the following tasks:

## Details
- Update the POST `/dictionary/register` endpoint to re try a migration.
- Implement the retry data migration based on main ticket

leoraba:
# Description
>Note: This PR is a [Part # 3](https://github.com/overture-stack/lyric/issues/177) of a series of a PRs to implement [Dictionary migration](https://github.com/overture-stack/lyric/issues/169). 

It enables the possibility to retry "failed" Dictionary migration by adding a `force` flag on the Dictionary Registration endpoint.

## Details:
- Updated `POST /dictionary/register` endpoint to accept an optional query param `force` to _Re-register the current active dictionary on the category and retries the data migration_.

## Related tickets:
- https://github.com/overture-stack/lyric/issues/177

## This PR depends on: 
- https://github.com/overture-stack/lyric/pull/200


apps/server/swagger/dictionary-api.yml
```diff
@@ -38,8 +46,8 @@
         $ref: '#/components/responses/BadRequest'
       401:
         $ref: '#/components/responses/UnauthorizedError'
-      404:
-        $ref: '#/components/responses/NotFound'
+      409:
+        $ref: '#/components/responses/StatusConflict'
```
leoraba, line 50: if registering the same Dictionary for the category then it throws a `409` error, unless the "force" flag is true.

apps/server/swagger/dictionary-api.yml
```diff
@@ -2,9 +2,17 @@
...
     tags:
       - Dictionary
+    parameters:
+      - name: force
+        description: Re-registers the current active dictionary on the category and retries the data migration.
+        in: query
+        required: false
+        schema:
+          type: boolean
+          default: false
```
joneubank, line 15: Can you add more detail about this behaviour? I don't know what re-registering a dictionary will do. Is this the same as initiating a migration? If the dictionary version is the same as the current dictionary will it run the migration with the intention of revalidating the data (should have no impact on the current data state)?

packages/data-provider/docs/dictionary-registration.md
```diff
@@ -108,6 +108,10 @@ A Category is uniquely identified by its case-sensitive `categoryName`.
 
 Category groups data that is related and shares the same data structure, for that reason, a category must be associated to a registered dictionary. Over time, if the dictionary requires an update, the category needs to be updates accordingly, See [Dictionary Migration](#dictionary-migration) for more details.
 
+If a category is already using the same dictionary name and version, registration returns `409 Conflict`.
+
+Set the `force` query parameter to `true` to allow re-registration and trigger a migration (or retry) to revalidate existing category data against that dictionary.
```
joneubank: Repeating [my comment](https://github.com/overture-stack/lyric/pull/201/changes#r3219699680) since this is the full documentation about registering dictionaries:

> Can you add more detail about this behaviour? I don't know what re-registering a dictionary will do. Is this the same as initiating a migration? If the dictionary version is the same as the current dictionary will it run the migration with the intention of revalidating the data (should have no impact on the current data state)?

packages/data-provider/src/services/migrationService.ts
```diff
@@ -214,7 +214,7 @@ const migrationService = (dependencies: BaseDependencies) => {
 
 				const newMigration: NewDictionaryMigration = {
 					categoryId,
-					fromDictionaryId,
+					fromDictionaryId: fromDictionaryId ?? toDictionaryId,
```
joneubank: This is confusing to read - perhaps we should leave it as required and the caller can indicate that `fromDictionaryId` value should be the same value as `toDictionaryId`. I would not expect the migration service to assume that if `from` value is not provided you reuse the `to` value.

packages/data-provider/src/services/dictionaryService.ts
```diff
@@ -135,7 +137,31 @@ const dictionaryService = (dependencies: BaseDependencies) => {
...
+					logger.error(LOG_MODULE, errorMessage);
+					throw new Error(errorMessage);
+				}
+
+				return { dictionary: savedDictionary, category: foundCategory, migrationId: resultMigration.data };
+			}
+
+			throw new StatusConflict(
+				`Category '${categoryName}' with Dictionary '${savedDictionary.name}' version '${savedDictionary.version}' already exists`,
+			);
```
joneubank: This is not very DRY, a lot of repeated code from the else block. The only difference is omitting the `fromDictionaryId` and changed log statements. Is there a way to write this that won't open us up to a future code change that is only made in one branch of this if statement?

apps/server/swagger/dictionary-api.yml
leoraba: description updated to `Runs dictionary registration and migration again for this category, even if the same dictionary is already registered. Use this only when a previous migration ended unexpectedly and you must rerun both steps.`

packages/data-provider/docs/dictionary-registration.md
leoraba: doc updated

packages/data-provider/src/services/migrationService.ts
leoraba: code updated to make `fromDictionaryId ` required

packages/data-provider/src/services/dictionaryService.ts
leoraba: refactored this code to make it DRY and reusing the initiateMigration logic.

leoraba:
PR updated to address feedback received. Ready for code review

joneubank:
Could you add a bit more description to the PR to explain "what happens if the force flag is used when migration is not in failed state"? For example, does it interrupt a migration thats in progress, does it repeat a completed migration. Any behaviours that are modified by this force flag need to be claerly stated.

packages/data-provider/docs/dictionary-registration.md
```diff
@@ -108,6 +108,8 @@ A Category is uniquely identified by its case-sensitive `categoryName`.
 
 Category groups data that is related and shares the same data structure, for that reason, a category must be associated to a registered dictionary. Over time, if the dictionary requires an update, the category needs to be updates accordingly, See [Dictionary Migration](#dictionary-migration) for more details.
 
+If a category is already using the same dictionary name and version, registration returns `409 Conflict` by default. If the `force` query parameter is set to `true`, Lyric retries only when the same dictionary is already registered and the latest migration previously failed. If no prior failed migration exists, the `force` flag is ignored.
```
leoraba, line 111: updated doc to describe `force` flag behaviour

packages/data-provider/src/services/dictionaryService.ts
```diff
@@ -131,11 +133,62 @@ const dictionaryService = (dependencies: BaseDependencies) => {
...
-			return { dictionary: savedDictionary, category: foundCategory };
+			// check last migration of this category to find if it failed, if it failed and forceRegistration is true,
+			// we will re-initiate the migration with the new dictionary
+			const activeMigration = await getMigrationsByCategoryId(foundCategory.id, { pageSize: 1, page: 1 });
+
+			if (
+				forceRegistration &&
+				activeMigration.result.length === 1 &&
+				activeMigration.result.at(0)?.status === 'FAILED'
+			) {
```
leoraba, line 173: a migration can be re-run only when last migration has `FAILED` and `force` flag is set to `true`

leoraba:
PR updated, changes on readme and openapi doc, to indicate `force` flag behaviour: `"If the 'force' query parameter is set to 'true', Lyric retries only when the same dictionary is already registered and the latest migration previously failed. If no prior failed migration exists, the 'force' flag is ignored."`

joneubank (approved):

PR #202, opened 2026-05-04
Dictionary Migration - Integration tests

leoraba:
# Description
This PR form parts of a series of PRs to implement [Dictionary migration](https://github.com/overture-stack/lyric/issues/169)

## Details
- Fix an issue running the integration test, tests have been broken since the implementation of the worker threads.
- Adding integration tests to cover Dictionary registration endpoint on a new category
- Adding integration tests to cover Dictionary registration endpoint on an existing category trigger Data Migration.

### Issue: 
- https://github.com/overture-stack/lyric/issues/203

## This PR depends on: 
- https://github.com/overture-stack/lyric/pull/200
- https://github.com/overture-stack/lyric/pull/201


packages/data-provider/test/integration/routers/submission/submissionRouter-submitFiles-persistence.spec.ts
```diff
@@ -9,6 +9,36 @@ import { createLyricProvider, type LyricProvider } from '../../dependencies/lyri
 import { createTestApp } from '../../dependencies/testServer.js';
 import { getContainers } from '../../globalSetup.js';
 
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
+const waitForSubmissionToStopValidating = async ({
```
leoraba, line 14: submission is being validated in a worker thread, which we can't control when it finished validating, adding this function to "wait" until the validation process completes

packages/data-provider/src/workers/workerPoolManager.ts
```diff
@@ -9,6 +8,29 @@ import type { CommitWorkerInput, DataValidationWorkerInput, WorkerFunctions, Wor
 
 const LOG_MODULE = 'WORKER_POOL_MANAGER';
 
+/**
+ * Resolves the appropriate worker file to use.
+ * For compiled/transpiled environments use .js, integration tests use .ts files directly, so the tsx loader
+ * is required for process workers to execute TypeScript files.
+ */
+const getWorkerPoolConfig = () => {
```
leoraba: This function fix running the integration tests, added a tsdoc with the explanation

packages/data-provider/test/integration/routers/dictionary/dictionaryMigration.spec.ts
```diff
@@ -0,0 +1,208 @@
+import { expect } from 'chai';
+import { after, afterEach, before, beforeEach, describe, it } from 'mocha';
+import supertest from 'supertest';
+
+import { dictionarySportsData } from '../../../fixtures/dictionarySchemasTestData.js';
+import { createLyricProvider } from '../../dependencies/lyricProvider.js';
+import { createTestApp } from '../../dependencies/testServer.js';
+import { getContainers } from '../../globalSetup.js';
+
+type LyricProvider = Awaited<ReturnType<typeof createLyricProvider>>;
```
demariadaniel: LyricProvider definition is duplicated across test files

demariadaniel: Same for a handful of test setup definitions, `VALID_CATEGORY_NAME`, `updatedSportSchema`, etc.

packages/data-provider/test/integration/routers/dictionary/dictionaryMigration.spec.ts
```diff
@@ -0,0 +1,208 @@
...
+
+	before(async () => {
+		schemaServiceUrl = getContainers().providerConfig.schemaService.url;
+		lyricProvider = await createLyricProvider(getContainers().providerConfig);
+		appDictionary = createTestApp(lyricProvider.routers.dictionary);
+		appMigration = createTestApp(lyricProvider.routers.migration);
+	});
+
+	beforeEach(async () => {
+		// processInsertRecordsAsyncStub.resetHistory();
```
JamesTLopez: Is this something to be added in the future? If so, lets add a todo or remove it.

demariadaniel: Remove commented code if unused

packages/data-provider/test/integration/routers/dictionary/dictionaryMigration.spec.ts
```diff
@@ -0,0 +1,208 @@
...
+		await seedDictionaryInSchemaService();
+		await seedDictionaryInSchemaService(VALID_DICTIONARY_NAME, NEW_DICTIONARY_VERSION, updatedSportSchema);
+	});
+
+	afterEach(async () => {
+		await getContainers().resetDatabases();
+	});
+
+	after(async () => {
+		// submissionProcessorFactory.create = originalCreate;
```
demariadaniel: Remove here as well

packages/data-provider/test/integration/routers/dictionary/dictionaryMigrationData.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+		);
+		expect(sourceDictionary).to.exist;
+
+		// save an invalid record (missing the required 'description' field in the new schema)
+		await lyricProvider.repositories.submittedData.save({
+			data: {
+				sport_id: '1',
+				name: 'Soccer',
+			},
+			dictionaryCategoryId: category!.id,
```
JamesTLopez: Pretty annoying that typescript doesn't understand chai's assertion here requiring the non null assertion...

Another way we could possibly type this is using typescripts asserts, but I think this is a good use-case to use the non null assertion to keep it as it is.

```ts
   function assertExists<T>(value: T | null | undefined): asserts value is T {
       expect(value).to.exist;
   }
   
```

packages/data-provider/test/integration/routers/dictionary/dictionaryRouter-registration.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+		lyricProvider = await createLyricProvider(getContainers().providerConfig);
+		app = createTestApp(lyricProvider.routers.dictionary);
+	});
+
+	afterEach(async () => {
+		await getContainers().resetDatabases();
+	});
+
+	after(async () => {
+		// submissionProcessorFactory.create = originalCreate;
```
demariadaniel: Remove unused comments

packages/data-provider/test/integration/routers/dictionary/dictionaryMigration.spec.ts
leoraba: Moved this type and constants to a common file `fixtures.ts` to be reused in other tests.

packages/data-provider/test/integration/routers/dictionary/dictionaryMigration.spec.ts
leoraba: removed commented code as it's not used anymore

packages/data-provider/test/integration/routers/dictionary/dictionaryMigration.spec.ts
leoraba: comment removed!

packages/data-provider/test/integration/routers/dictionary/dictionaryRouter-registration.spec.ts
leoraba: removed

packages/data-provider/test/integration/routers/dictionary/dictionaryMigration.spec.ts
leoraba: comment removed

packages/data-provider/test/integration/routers/dictionary/dictionaryMigrationData.spec.ts
leoraba: thanks for the suggestion! code has been updated to use this function and fix the typescript issue

leoraba:
PR updated based on feedback received, it is now ready for code review.

demariadaniel (approved):
Thanks for the cleanup, looking good to me

JamesTLopez (approved):
🔥

packages/data-provider/src/workers/workerPoolManager.ts
joneubank: Glad you were able to solve this problem and allow workers to run in the runtime and testing contexts.

I am concerned about putting these testing focused changes into code that is included in the exported library. This will be imported by everything downstream and included in the runtime code even though we wouldn't ever want it to run. I would suggest then that there is another way to update the code architecture to alow this configuration to be provided by the testing code instead of kept in our library code.

Perhaps the `createWorkerPool` function can take a `WorkerPoolConfig` object as an optional parameter, and default to the original values, but in the test case we can specify the properties to be used? That might not describe a complete solution but might give a path towards one.

packages/data-provider/test/integration/assertions.ts
joneubank, line 1: Cool, I like this utility.

joneubank (commented):
Looks good. If you have a few minutes to spare can you look at reorganizing the worker pool config to not be encoded in the runtime source, instead this should be configurable and allow the test code to provide the alternate values.

packages/data-provider/test/integration/dependencies/lyricProvider.ts
```diff
@@ -24,39 +26,49 @@ export type LyricProviderConfig = {
...
+const testWorkerPoolConfigResolver: WorkerPoolConfigResolver = () => ({
+	workerPath: fileURLToPath(new URL('../../../src/workers/workerpool.ts', import.meta.url)),
+	poolOptions: {
+		workerType: 'process',
+		forkOpts: {
+			execArgv: ['--import=tsx'],
+		},
+	},
+});
+
```
leoraba, line 56: this function is required to run integration tests only as it looks for the `.ts` file.

packages/data-provider/src/core/provider.ts
```diff
@@ -37,14 +37,19 @@ import * as schemaUtils from '../utils/schemas.js';
...
+};
 
 /**
  * The main provider of submission resources
  * @param configData Environment variables required to configure resources
+ * @param options Optional provider-level overrides (mainly for test wiring)
  * @returns A provider to get access to resources
  */
-const provider = (configData: AppConfig) => {
+const provider = (configData: AppConfig, options?: ProviderOptions) => {
```
leoraba, line 52: Lyric provider accepts a new optional parameter, it's actually only used to run integration tests.

packages/data-provider/src/workers/workerPoolManager.ts
leoraba: Updated this code, now Lyric provider accepts an optional parameter, which is used to resolve worker `.ts` file during integration tests.

leoraba:
PR updated to resolve feedback received, is now ready for code review.

PR #208, opened 2026-06-26
feat: Publish committed records to Kafka document topic

justincorrigible:
## Summary

Add Kafka publishing to Lyric: after each commit, affected records are pushed as individual JSON messages to a configurable Kafka topic. A `published_at` timestamp on the `submissions` table now tracks successful publication without touching submission status or the existing state machine.

## Issues

No ticket. needed for overture-dev setup.

## Description of Changes

This PR adds opt-in Kafka publishing. Setting `KAFKA_BROKERS` enables it; omitting it leaves Lyric behaviour unchanged.

**_NOTE_:** `onFinishCommit`'s type signature changed, and may require adjustment when integrating.

### `packages/data-model`

- Added nullable `published_at` timestamp column to the `submissions` table
- Migration 0013: `ALTER TABLE submissions ADD COLUMN published_at timestamp`; backfills the Unix epoch (`1970-01-01`) for all existing `COMMITTED` rows as a "pre-Kafka" sentinel (distinguishes "committed before Kafka existed" from "publish failed")

### `packages/data-provider`

- **Breaking change:** `onFinishCommit` callback signature changed from `() => void` to `(resultOnCommit: ResultOnCommit) => Promise<void>` - coordinated with PCGL/iMicroSeq (@leoraba)
- New `src/external/kafkaPublisher.ts`: `createKafkaPublisher` factory implementing the document topic pattern - one message per affected record (inserts, updates, deletes), batched in a single `producer.send` call; deleted records have `isValid` forced to `false`
- New `src/external/kafkaPublishTracker.ts`: `createPublishTracker` writes `published_at` timestamp on successful send; publish and tracking errors are isolated in separate try/catch blocks
- `provider()` accepts an optional `db` parameter so the server can share one DB pool between the provider and the publish tracker
- New exports: `createKafkaPublisher`, `createPublishTracker`, `KafkaProducer`, `KafkaPublisherConfig`, `connect`, `getLogger`, `Logger`
- 17 new mocha/chai unit tests for `createKafkaPublisher` in `test/unit/external/kafkaPublisher.spec.ts`
- New `docs/kafka-publishing.md`: integration guide for library consumers (message format, wiring pattern, config reference, error handling, publish tracking semantics)

### `apps/server`

- Added `kafkajs` dependency
- New `src/config/kafka.ts`: `setupKafka` reads env vars, connects producer, checks/creates the topic via Admin client on startup, wires `createKafkaPublisher` + `createPublishTracker`; kafkajs retry config (5 retries, 300ms-30s exponential backoff); graceful disconnect
- `src/config/app.ts`: refactored from a static `appConfig` export to `buildAppConfig(overrides)` + `getDbConfig()` for composable setup that accepts the Kafka callback
- `src/server.ts`: creates a shared logger and DB pool before the provider; passes both to `setupKafka` and `provider`; Kafka disconnect included in graceful shutdown sequence before worker pool drain
- `apps/server/.env.schema`: added `KAFKA_BROKERS`, `KAFKA_TOPIC`, `KAFKA_CLIENT_ID` (all commented - opt-in)

### Special Instructions

- Run `pnpm install` to install `kafkajs`
- Run migration 0013 (`pnpm run migrate-dev`)
- Kafka is **opt-in**: set `KAFKA_BROKERS` to enable; omit it to run unchanged
- New env vars (all optional when Kafka is not used): `KAFKA_BROKERS`, `KAFKA_TOPIC`, `KAFKA_CLIENT_ID` - see `DEVELOPMENT.md`

## Readiness Checklist

- [x] **Self Review**
- [ ] **Formatting**
  - Automated code formatters (ie. Prettier) have been run
- [x] **Local Testing**
- [x] **Updated Tests**
  - 17 new unit tests added for `createKafkaPublisher` covering the publish path, `onSuccess` callback, `onError` callback, and empty result short-circuit
- [x] Documentation
  - [x] New environment variables added to `apps/server/.env.schema` and documented in `DEVELOPMENT.md`
  - [ ] All changes to server HTTP endpoints have open-api documentation (no new endpoints in this PR)
  - [x] All new functions exported from their module have TSDoc comment documentation


packages/data-model/src/models/submissions.ts
```diff
@@ -79,6 +79,7 @@ export const submissions = pgTable(
 			.notNull(),
 		errors: jsonb('errors').$type<SubmissionErrors>(),
 		organization: varchar('organization').notNull(),
+		publishedAt: timestamp('published_at'),
```
leoraba: not really sure this field is required, a submission once `COMMITTED` it cannot be modified, so the last `updated_at` date will be same as `published_at`

justincorrigible: this new column is meant to help distinguish records that were submitted and committed, and have been shared externally e.g. kafka responds with 200. this is particularly useful in case of needing to suppress data, and also when lyric may crash for whatever reason, we'll be able to resume publishing after only the remaining unpublished records. a future endpoint to automate that is in the roadmap now.

side note: originally thought "indexed_at" which may have been more (seemingly) declarative, but other platforms may instead send the records into a data lake, an aggregator, or some other form of big data management. beyond that, added a new column so that the existing records' values won't need to be modified... thought of adding a PUBLISHED state, but that'd probably complicate things unnecessarily.

this column is unused as of right now, and it's more to capture "publishing" info in case we do want to proceed in the direction I suggest. TBD

justincorrigible:
pushed an update because I realized the package I had used wasn't compatible with the latest versions of `Kafka` we're deploying in `overture-dev`. the new one supports v3.x (used in other envs) and v4.x (the current one).

note: had some issues with the package in the build pipeline because of the way we're blocking the post install scripts. the dockerfile acrobatics are an isolated solution for it without compromising the rest of the image

packages/data-model/migrations/meta/0014_snapshot.json
leoraba, line 1: this snapshot is not correct, it doesn't match with the current db schema, and it needs to be re-generated.
1. back up `0014_add_published_at_to_submissions.sql`
2. Delete all 3 artifacts: `0014_add_published_at_to_submissions.sql`, `0014_snapshot.json` and the idx 14 in meta/_journal.json
3. Regenerate using `pnpm --filter @overture-stack/lyric-data-model build:all` 
4. Rename the generated files: 
 - The 0014_*.sql to `0014_add_published_at_to_submissions.sql`
 - The journey entry tag to `0014_add_published_at_to_submissions`
5. Replace the generated .sql content with the backed-up original

apps/server/src/config/kafka.ts
```diff
@@ -0,0 +1,75 @@
...
+			retry: {
+				initialRetryTime: 300,
+				maxRetryTime: 30_000,
+				retries: 5,
+			},
+		},
+	});
+
+	logger.info(`[kafka] Connecting to broker(s): ${brokers}`);
+	await producer.connect();
```
leoraba: when it tries to connect and fails, application enters in an infinite loop, need to add a stopper when connection fails and stop the application

pnpm-workspace.yaml
```diff
@@ -4,6 +4,7 @@ packages:
   # all packages in subdirs of packages/
   - 'packages/*'
 allowBuilds:
+  '@confluentinc/kafka-javascript': true
```
leoraba, line 7: is it really needed to be `true`?

justincorrigible: Yes, because of the way our build pipeline works: `@confluentinc/kafka-JavaScript` bundles a native add-on built on `librdkafka` (a C library). It uses `node-pre-gyp` in its install lifecycle script to download a pre-built platform binary (the .node file). Without that binary, the package fails at runtime e.g. once deployed, with "Could not locate the bindings file" as a critical error.

The cause is that `pnpm` workspaces block all package lifecycle scripts by default, even on a plain `pnpm install` with no flags (which is a security feature we want). Now, to state the obvious, to benefit those who don't know, `allowBuilds` is the workspace-level opt-in list for which packages are permitted to run those scripts.... this should be used sparingly, and it works as an exception to that security feature, which is why @leoraba is asking to confirm.
Without `'@confluentinc/kafka-javascript': true`, pnpm silently skips its install script and the binary never lands, breaking both local dev and even the CI build stage (which does a full install before compilation).

The Dockerfile prod-deps stage is a separate concern: it uses --no-scripts explicitly as an OWASP requirement, which overrides `allowBuilds` entirely; so there we have to invoke `node-pre-gyp` manually after the fact.
i.e. `allowBuilds` only matters for installs that don't pass --no-scripts.

Hope that clarifies why this line is here.

packages/data-model/migrations/meta/0014_snapshot.json
justincorrigible: this should be resolved in the latest commits. resolving for now, but please reopen here if needed.

packages/data-provider/docs/kafka-publishing.md
```diff
@@ -0,0 +1,81 @@
...
+Each message value is a JSON string:
+
+```json
+{
+  "data": { "field": "value" },
+  "entityName": "donor",
+  "isValid": true,
+  "organization": "EXAMPLE-ORG",
+  "systemId": "ABC-12345"
+}
```
leoraba, line 23: This message format matches current Maestro expectations, which might be a limitation for others (hypothetically) Kafka consumers, for example, a message doesn't have submission date, published date, action (new, update, delete), user, etc.  
Open for a discussion if this is a requirement for now or another PR.
@joneubank

leoraba (approved):
Approved, just left a comment regarding the message format, which might need an improvement later

packages/data-provider/src/external/kafkaPublisher.ts
```diff
@@ -0,0 +1,66 @@
...
+	/** Called when `producer.send` throws after all retries. Defaults to `console.error`. */
+	onError?: (err: unknown) => void;
+	/** Called after a successful send; isolated from publish errors. Use {@link createPublishTracker} to persist `published_at`. */
+	onSuccess?: (submissionId: number) => Promise<void>;
+	producer: KafkaProducer;
+	/** Topic to publish to; created automatically on startup if it does not exist. */
+	topic: string;
+};
+
+const toMessage = (record: SubmittedDataResponse, action: 'delete' | 'insert' | 'update') => ({
```
joneubank: Surprised we don't have an existing type that enumerates these options. Would prefer these be defined as constants with a corresponding type. This is reflective of the type options that already exist, so it might exist somewhere

packages/data-provider/src/workers/workerPoolManager.ts
```diff
@@ -97,7 +97,7 @@ export const createWorkerPool = (configData: AppConfig, options?: CreateWorkerPo
 				const resultCommit = await proxy.commitSubmission(input);
 
 				if (configData.onFinishCommit && resultCommit) {
-					configData.onFinishCommit(resultCommit);
+					await configData.onFinishCommit(resultCommit);
```
joneubank, line 104: @leoraba is this a case where we intended the onFinishCommit to occur asynchronously and not be waited for? or is this a missed await? if the former, needs a comment to clarify why its not being waited for.

justincorrigible: submissionService.ts:99 calls `workerPool.commitSubmission(commitData)` with no await: as the worker runs fully async and the HTTP 200 goes back immediately with PROCESSING status... so the `await` here is meant to just control the sequencing within the worker's own flow. 

may be worth just adding a comment to make the intent obvious 🤷 (adding one now, for good measure)

joneubank: I've reviewed the context - onFinishCommit likely does not *need* to be awaited, but this is happening in a worker already so we're not blocking responses to a user so its safer to await so that our error handling can handle anything unexpected in this function call.

tldr - await is good ✅

apps/server/src/server.ts
```diff
@@ -3,15 +3,22 @@ import express from 'express';
...
 import healthRouter from './routes/health.js';
 import pingRouter from './routes/ping.js';
 
 const { allowedOrigins, port, corsEnabled } = getServerConfig();
 
+const logger = getLogger({ level: process.env.LOG_LEVEL || 'info' });
+const db = connect(getDbConfig());
+const kafka = await setupKafka(db, logger);
+
+const appConfig = buildAppConfig({ onFinishCommit: kafka?.onFinishCommit });
```
joneubank, line 20: I think we are meant to pass our logger into the appConfig, so that the lyric provider uses our logger for all its logs.

(Not needed for this PR, just places for improvement).

apps/server/src/config/kafka.ts
```diff
@@ -0,0 +1,85 @@
...
+} from '@overture-stack/lyric';
+
+type LyricDb = ReturnType<typeof connect>;
+
+type KafkaSetupResult = {
+	disconnect: () => Promise<void>;
+	onFinishCommit: (result: ResultOnCommit) => Promise<void>;
+};
+
+export const setupKafka = async (db: LyricDb, logger: Logger): Promise<KafkaSetupResult | undefined> => {
```
joneubank: I'm surprised our kafka service needs to read from the DB instead of having all the required data for its messages passed to it.

Edit: just read through the kafka producer... we are recording a published_at property. Kafka is writing to the DB, not reading from it. This might be violation of the single responsibility principle, kafka should request of a data service to mark the record as published instead of sending the db requests directly?

packages/data-model/src/models/submissions.ts
joneubank: Agreed that `published_at` corresponds to the kafka action that is used to set this, not `indexed_at`.

I think that I agree with Leo that tracking publishing is not the responsibility of Lyric. If a publishing service needs to fix the published state, they'll need to manage their own state instead of relying on this column in lyric.

apps/server/src/config/kafka.ts
```diff
@@ -0,0 +1,85 @@
...
+	});
+
+	logger.info(`[kafka] Connecting to broker(s): ${brokers}`);
+	try {
+		await producer.connect();
+	} catch (err) {
+		logger.error('[kafka] Failed to connect to broker(s) after retries, shutting down', err);
+		throw err;
+	}
+	logger.info(`[kafka] Connected (clientId: ${clientId}). Publishing commits to topic: ${topic}`);
```
joneubank, line 72: :+1: was confirming that we log the clientId, since its important that the logs say it if we end up using the default.

joneubank, line 72: :+1: was confirming that we log the clientId, since its important that the logs say it if we end up using the default.

apps/server/src/config/kafka.ts
```diff
@@ -0,0 +1,85 @@
...
+	disconnect: () => Promise<void>;
+	onFinishCommit: (result: ResultOnCommit) => Promise<void>;
+};
+
+export const setupKafka = async (db: LyricDb, logger: Logger): Promise<KafkaSetupResult | undefined> => {
+	const brokers = process.env.KAFKA_BROKERS;
+	const clientId = process.env.KAFKA_CLIENT_ID ?? 'lyric';
+	const topic = process.env.KAFKA_TOPIC;
+
+	if (!brokers) return undefined;
```
joneubank: Do we want to log a warning if the user provides `KAFKA_CLIENT_ID` or `KAFKA_TOPIC` env params but misses the kafka-brokers? indicates a potential misconfiguration.

packages/data-model/src/models/submissions.ts
justincorrigible: 2 vs 1 is majority. removing this without replacement for now

apps/server/src/config/kafka.ts
justincorrigible: good call! adding that

apps/server/src/server.ts
justincorrigible: added as tech debt in the latest commit, so it doesn't fall through the cracks in the future. marking as resolved here, because scope

apps/server/src/config/kafka.ts
justincorrigible: marking comment as resolved because it's duplicated

apps/server/src/config/kafka.ts
justincorrigible: removed that whole part, so this should not be necessary

packages/data-provider/src/external/kafkaPublisher.ts
justincorrigible: Adding a new type for this, because the only ones that already exist are:

`SUBMISSION_ACTION_TYPE`: `['INSERTS', 'UPDATES', 'DELETES']` which are plural, uppercase, and represent submission categories, not individual record events.
`AUDIT_ACTION`: `['UPDATE', 'DELETE', 'MIGRATION']` with no insert, includes MIGRATION (?), and "audit" seems like it would diverge over time.

joneubank (approved):

apps/server/src/config/app.ts
```diff
@@ -8,9 +8,10 @@ const DEFAULT_SUBMISSION_MAX_SIZE = 10 * 1024 * 1024; // 10Mb
 
 export const getServerConfig = () => {
 	return {
-		port: process.env.PORT || 3030,
 		allowedOrigins: process.env.ALLOWED_ORIGINS?.split(',') || [],
 		corsEnabled: getBoolean(process.env.CORS_ENABLED, false),
+		logLevel: process.env.LOG_LEVEL || 'info',
+		port: process.env.PORT || 3030,
```
justincorrigible, line 14: alphabetics are bae 👍

packages/data-provider/docs/kafka-publishing.md
justincorrigible: documenting this as a roadmap item, so it doesn't fall through the cracks. marking as resolved here

Issue #171, opened 2025-11-28
Request - Change `protectedMethods` configuration


leoraba:
## Summary of request
A Request to remove `protectedMethods` auth configuration in Lyric provider, and add new `publicRead` (default `false`).

## Details
- Remove `protectedMethods` from Lyric provider configuration
- Add `publicRead` in Lyric provider configuration (default `false`). 
  - On `false` it should protect every endpoint
  - on `true` it should open endpoints that reads data. (example: fetch submitted Data)
- `publicRead` should be ignored when  `auth_enabled=false`
- Update documentation accordingly

PR #209, opened 2026-07-03
#171: Refactor Submission Record data access layer for new schema

leoraba:
## Summary
Refactors the functions responsible for fetching, updating, and deleting Submission Records to align with the new database structure. This updates the data access logic without changing the external behaviour of these operations.

## Issues
- [#171](https://github.com/Pan-Canadian-Genome-Library/Roadmap/issues/171)

## Description of Changes
- Removed the legacy columns for `data` and `errors` in Submission table. **NOTE: This permanently deletes the data and errors values stored under the old schema. This data has already been migrated to the new schema, so no information is lost.**
- Added 2 new repository files: **submissionFilesRepository.ts** and **submissionRecordsRepository.ts** to access submission files and records.
- Updated **Submission service** to adapt to new database schema to fetch submission records 

## Readiness Checklist

- [ ] **Self Review**
- [ ] **Formatting**
  - Autmated code formatters (ie. Prettier) have been run
- [ ] **Local Testing**
- [ ] **Updated Tests**
- [ ] Documentation


apps/server/swagger/submission-api.yml
```diff
@@ -98,7 +98,7 @@
       503:
         $ref: '#/components/responses/ServiceUnavailableError'
 
-/submission/{submissionId}/{actionType}:
+/submission/{submissionId}/data:
```
leoraba: renamed this endpoint to `/data` and changed to query params to include `recordId` and `fileId`

apps/server/swagger/submission-api.yml
```diff
@@ -56,7 +56,7 @@
 
 /submission/{submissionId}/details:
   get:
-    summary: Fetch Submission Data records. Sorted in their original file order and grouped by `inserts`, `updates`, and `deletes`.
+    summary: Fetch Submission Data records. Sorted in their original file order and grouped by `insert`, `update`, and `delete`.
```
leoraba, line 59: Changed the values of this query param to be consistent through all the application

packages/data-model/migrations/0017_remove_submission_data.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE "submissions" DROP COLUMN IF EXISTS "data";--> statement-breakpoint
+ALTER TABLE "submissions" DROP COLUMN IF EXISTS "errors";
```
leoraba, line 2: removing data and errors column from submission table, this PR assumes the SQL script migration (0016) moved all submissions data into it's own separate table

packages/data-provider/src/services/submission/submissionProcessor.ts
```diff
@@ -739,35 +773,52 @@ const createSubmissionProcessor = (dependencies: BaseDependencies) => {
...
 				updatedBy: username,
 				status: 'VALIDATING',
 			});
 
+			const insertRecords = parseRecordsToInsert(records, schemasDictionary);
+
+			await Promise.all(
+				Object.entries(insertRecords).map(([entityName, entityRecords]) =>
+					dependencies.db.transaction(async (tx) => {
+						const savedFileId = await submissionFilesRepository.save(
```
leoraba, line 800: wrapped in a db transaction storing each file along with it's records for atomicity.

packages/data-provider/src/repository/activeSubmissionRepository.ts
```diff
@@ -11,33 +11,24 @@ import { inProcessSubmissionStatus, openSubmissionStatus } from '../utils/submis
...
 		updatedBy: true,
-	};
-
-	// Submission columns for full detail queries including `data` and `errors` columns
-	const submissionColumnsWithData: BooleanTrueObject = {
-		...submissionColumns,
-		data: true,
-		errors: true,
-	};
+	} as const satisfies PartialColumns<Omit<Submission, 'dictionaryCategoryId' | 'dictionaryId'>>;
```
joneubank, line 31: 👍

packages/data-provider/src/services/submission/submissionProcessor.ts
```diff
@@ -677,32 +666,77 @@ const createSubmissionProcessor = (dependencies: BaseDependencies) => {
 			// Aggegates all Update changes on Submission
 			// Note: We do not include records involving primary ID fields changes in here. We would rather do a DELETE and an INSERT
 			const updatedActiveSubmissionData: Record<string, SubmissionUpdateData[]> = mergeUpdatesBySystemId(
-				submission.data.updates ?? {},
+				// formattedSubmissionRecordsToUpdate,
```
joneubank, line 669: Can we remove commented code?

packages/data-provider/src/repository/submissionRecordsRepository.ts
```diff
@@ -0,0 +1,368 @@
...
+
+	const saveMany = async (
+		inputs: NewSubmissionRecord[],
+		tx?: PgTransaction<PostgresJsQueryResultHKT, SubmissionRecord, ExtractTablesWithRelations<SubmissionRecord>>,
+	): Promise<number[]> => {
+		if (!inputs.length) {
+			return [];
+		}
+		try {
+			// TODO: Insert in batches if inputs.length > 1000 to avoid exceeding the maximum number of parameters in a single query
```
joneubank, line 76: Should we just add this batching now before committing this PR?

packages/data-provider/src/repository/submissionRecordsRepository.ts
```diff
@@ -0,0 +1,368 @@
...
+		} catch (error) {
+			logger.error(LOG_MODULE, `Failed deleting Submission Records by fileIds`, error);
+			throw new ServiceUnavailable();
+		}
+	};
+
+	return {
+		saveMany,
+
+		saveManyForFile: async (
```
joneubank, line 112: Inconsistent code here. I would recommend continuing the pattern of defining the functions above and then returning them by name instead of having some defined above and some defined inline.

packages/data-provider/src/repository/submissionRecordsRepository.ts
```diff
@@ -0,0 +1,368 @@
...
+		 * It can update records to 'VALID', 'RECEIVED', or 'INVALID' states, and also set errors for invalid records.
+		 * @param params
+		 * @param tx
+		 * @returns
+		 */
+		updateValidationState: async (
+			params: {
+				validRecordIds?: number[];
+				receivedRecordIds?: number[];
+				invalidRecords?: Array<{ id: number; errors?: SubmissionRecordError[] }>;
```
joneubank, line 236: Mixed use of `Array<X>` and `X[]`. The typescript preffered convention is to use `X[]` when defining arrays, but both are fine. Lets just be consistent here:

```suggestion
				receivedRecordIds?: number[];
				invalidRecords?: { id: number; errors?: SubmissionRecordError[] }[];
```

packages/data-provider/src/repository/submissionRecordsRepository.ts
```diff
@@ -0,0 +1,368 @@
...
+		},
+
+		/**
+		 * This function updates the validation state of submission records based on the provided parameters.
+		 * It can update records to 'VALID', 'RECEIVED', or 'INVALID' states, and also set errors for invalid records.
+		 * @param params
+		 * @param tx
+		 * @returns
+		 */
+		updateValidationState: async (
```
joneubank, line 232: - great to see TSDoc comment blocks
- best practice is to add them on all the functions here
- don't need to declare the `@ param` and `@ returns` data unless there is a meaningful description we need to add, TS provides the official function signature
- "This function" is not needed, we know this comment is for this function
- for a function named `updateValidationState`, the description"updates the validation state" is not particularly descriptive. The second sentence adds good character. How about something like "Sets the `validationState` property for many records at once. Provide record IDs grouped into arrays for `VALID`, `RECEIVED`, and `INVALID` records, and each record will be updated according to the array it is found in." I would then also describe the behaviour for what happens if an array is not provided (optional arrays), left empty, and what happens if the same ID is in multiple arrays. Are there any thrown error cases that the caller should be aware of?

packages/data-provider/src/services/submission/submissionProcessor.ts
```diff
@@ -441,7 +441,7 @@ const createSubmissionProcessor = (dependencies: BaseDependencies) => {
 
 				resultCommit.deletes.push({
 					data,
-					entityName,
+					entityName: '', // TODO: need to fetch the entityName
```
joneubank, line 444: Is this a TODO that should be resolved before this merges? Seems like we can't return this data correctly as this is written.

packages/data-provider/src/services/submission/submissionProcessor.ts
```diff
@@ -523,10 +523,31 @@ const createSubmissionProcessor = (dependencies: BaseDependencies) => {
...
+		const submissionRecords = await submissionRecordsRepository.getBySubmissionId(submissionId);
+
+		// Detect records where the same systemId has both an UPDATE and a DELETE staged, before
+		// running dictionary validation. Both sides of a conflict are rejected explicitly instead
+		// of letting one action silently win.
+		const conflictErrors = findUpdateDeleteConflicts(submissionRecords);
+		const conflictingRecordIds = extractRecordIdsFromSubmissionErrors(conflictErrors);
+
+		if (conflictingRecordIds.size > 0) {
+			logger.error(
```
joneubank, line 535: This case is not an "error" as far as our system is concerned, this is a normal situation that we are prepared to handle, and should be logged as debug comment or info at most. This is a data conflict that the user will be told about and can resolve without any system alerting being required.

packages/data-provider/src/services/submission/submissionService.ts
```diff
@@ -147,24 +149,28 @@ const submissionService = (dependencies: BaseDependencies) => {
...
-	 * @param {number} submissionId
-	 * @param {string} entityName
-	 * @param {string} username
+	 * @param {number} submissionId - Submission ID
+	 * @param {string} username - User name performing the action
+	 * @param {object} filter - Filter to identify the entity to be removed
 	 * @returns { Promise<SubmitDataResult>}
 	 */
-	const deleteActiveSubmissionEntity = async (
+	const deleteByRecordIdOrFileId = async (
```
joneubank, line 165: This TSDoc comment is rendered by most IDEs using a markdown parser. This means a single new line character is ignored (assumed to be formatting for file width and not starting a new paragraph). As a consequence, this documentation comment renders as follows, with all these sentence fragments runnign together as a single paragraph:

Image

I would request you to separate all these points by either (a) using full sentences with periods to indicate the end of the sentence, (b) using two new lines (visible blank line) between paragraphs to separate them, or (c) use a bulleted list (new lines starting like ` - text` are treated as a bulleted list).

packages/data-provider/src/services/submission/submissionService.ts
```diff
@@ -253,17 +270,30 @@ const submissionService = (dependencies: BaseDependencies) => {
...
+			recordsPaginated.map(async (response) => {
+				const submissionRecordsSummary = await submissionRecordsRepository.getRecordsSummaryBySubmissionId(response.id);
+				const formattedDataSummary = buildDataSummary(submissionRecordsSummary);
+
+				return createSubmissionSummaryResponse({
+					...response,
+					data: formattedDataSummary,
+				});
+			}),
+		);
```
joneubank, line 284: This seems to be working well, but some non-required suggestions here to improve code readability:

- `createSubmissionSummaryResponse` name is confusing, this is actually creating a summary of one Record in a submission, formatted for an API response. This name doesn't mention this is a single record so it reads like this function creates a summary for the entire Submission.
- the desired type of the result is ambiguous. It is clear that the inferred type is `ResultType<typeof createSubmissionSummaryResponse>[]`, but I do not know what that type is or why you want it. The name of the variable `result` helps me understand it is used as the result in our function output, but we could also make the typing explicit for readers and also provide type hints and corrections to a developer at the return statement. There are two different ways to do this, I think I would choose the first suggestion here since it ensures that `result` variable has the needed type to be used in the function output:
  1. Declare the type for result: `const result: SubmissionSummaryResponse[] = ...`
  2. Declare the return type of the map: `recordsPaginated.map<SubmissionSummaryResponse>(...`

packages/data-provider/src/services/submission/submissionService.ts
```diff
@@ -253,17 +270,30 @@ const submissionService = (dependencies: BaseDependencies) => {
...
+				totalRecords: totalSubmissions,
 			},
-			result: recordsPaginated.map((response) => createSubmissionSummaryResponse(response)),
+			result,
 		};
 	};
 
 	/**
 	 * Get Submission by Submission ID
+	 * Returns the submission general information and includes the summary of the data and errors
```
joneubank, line 296: Repeated issue with TSDoc comment and new line formatting. I won't point out others, but this is a recurring issue.

packages/data-provider/src/services/submission/submissionService.ts
```diff
@@ -294,33 +330,45 @@ const submissionService = (dependencies: BaseDependencies) => {
...
+		const dictionary = await dictionaryRepository.getDictionary(
+			submission.dictionary.name,
+			submission.dictionary.version,
+		);
+
+		if (!dictionary) {
+			throw new InternalServerError(
+				`Dictionary '${submission.dictionary.name}' version '${submission.dictionary.version}' not found`,
+			);
+		}
```
joneubank, line 349: Would be useful for the function TSDoc documentation to describe the error conditions and what thrown values to expect.

packages/data-provider/src/repository/submissionRecordsRepository.ts
```diff
@@ -0,0 +1,368 @@
...
+
+import { BaseDependencies } from '../config/config.js';
+import { ServiceUnavailable } from '../utils/errors.js';
+import type { PaginationOptions, SubmissionRecordActionType, SubmissionRecordState } from '../utils/types.js';
+
+// This is the information stored about each individual submission record in the database, including it's entity name.
+export type SubmissionRecordWithEntityName = SubmissionRecord & { entityName: string };
+
+// Raw data returned from the database
+export type RecordsSummaryRepository = {
```
joneubank, line 22: I hate what I'm about to write, but: I think the correct name here is `SubmissionRecordRepositoryRecordSummary` since it is a Summary of a single Record from the repository for `SubmissionRecord`s. Using a naming pattern consistent with the other repositories is important, but because this name will be awkward I recommend adding descriptive TSDoc. The current name cannot be right though, this is not a Repository for  Record Summaries (implied by current name).

packages/data-provider/src/repository/submissionRecordsRepository.ts
```diff
@@ -0,0 +1,368 @@
...
+			return savedSubmissionRecords.map((record) => record.id);
+		} catch (error) {
+			logger.error(LOG_MODULE, `Failed saving '${inputs.length}' Submission Record records`, error);
+			throw new ServiceUnavailable();
+		}
+	};
+
+	const deleteByFileIds = async (
+		fileIds: number[],
+		tx?: PgTransaction<PostgresJsQueryResultHKT, SubmissionRecord, ExtractTablesWithRelations<SubmissionRecord>>,
```
joneubank, line 91: Should we write a type utility for `RepositoryTransaction<RecordSchema>`?

```ts
export type RepositoryTransaction<TSchema extends Record<string, unknown>> = PgTransaction<PostgresJsQueryResultHKT, TSchema, ExtractTablesWithRelations<TSchema>>;
```

So we can clean up these function signatures to be like:

```ts
const deleteByFileIds = async (
		fileIds: number[],
		tx?: RepositoryTransaction<SubmissionRecord>,
```

packages/data-provider/src/repository/submissionRecordsRepository.ts
```diff
@@ -0,0 +1,368 @@
...
+		if (!fileIds.length) {
+			return 0;
+		}
+		try {
+			const deletedRecords = await (tx || db)
+				.delete(submissionRecords)
+				.where(inArray(submissionRecords.fileId, fileIds))
+				.returning({ id: submissionRecords.id });
+			logger.info(LOG_MODULE, `Deleted '${deletedRecords.length}' Submission Record records by fileIds`);
+			return deletedRecords.length;
```
joneubank, line 102: Why do we return the length anyways?

packages/data-provider/src/services/submission/submissionProcessor.ts
```diff
@@ -677,32 +666,77 @@ const createSubmissionProcessor = (dependencies: BaseDependencies) => {
...
+			// Updating the Submission with the new data and 'VALIDATING' status before validation starts
+			await dependencies.db.transaction(async (tx) => {
+				await update(
+					submission.id,
+					{
+						updatedBy: username,
+						status: 'VALIDATING',
+					},
+					tx,
+				);
```
joneubank, line 686: For readability, we probably want to keep this update close to where we start the validation. Could we move this statement to the end of our list of updates so it comes directly before:

```ts
			// Perform Schema Data validation in a worker thread
			dependencies.workerPool.dataValidation({ submissionId: submission.id });
```

In fact, we may want to wrap both of those actions into a single function so that this process can just make a statement like `startAsyncValidation(submission.id, tx)` and we have confidence that the status will be set before the worker is added to the pool.

packages/data-provider/src/services/submission/submissionProcessor.ts
```diff
@@ -739,35 +773,52 @@ const createSubmissionProcessor = (dependencies: BaseDependencies) => {
...
+								state: 'RECEIVED',
+							})),
+							tx,
+						);
+					}),
+				),
+			);
+
 			// Perform Schema Data validation in a worker thread
 			dependencies.workerPool.dataValidation({ submissionId: activeSubmission.id });
```
joneubank, line 823: This process also starts data validation, but I don't see the status of the submission being set to `VALIDATING`. Is this correct behaviour?

packages/data-provider/src/services/submission/submissionProcessor.ts
```diff
@@ -795,43 +848,72 @@ const createSubmissionProcessor = (dependencies: BaseDependencies) => {
...
+			);
+			return updatedActiveSubmissionId;
 		});
 
-		logger.info(
-			LOG_MODULE,
-			`Updated Active submission '${updatedActiveSubmissionId}' with status '${newStatusSubmission}'`,
-		);
-		return updatedActiveSubmissionId;
+		return 0;
```
joneubank, line 894: I don't understand why we are returning a literal value here when the documentation states this will return with the updated submissionID. This behaviour either needs to be documented and justified or changed.

packages/data-provider/src/services/submission/submissionProcessor.ts
```diff
@@ -844,65 +926,72 @@ const createSubmissionProcessor = (dependencies: BaseDependencies) => {
...
+			await dependencies.db.transaction(async (tx) => {
+				// Updating the Submission with the new data and 'VALIDATING' status before validation starts
+				await submissionRepository.update(
+					submissionId,
+					{
+						updatedBy: username,
+						status: 'VALIDATING',
+					},
+					tx,
+				);
```
joneubank, line 952: Repeated comment about combining VALIDATING status update with the workerPool dispatch.

packages/data-provider/src/utils/submissionUtils.spec.ts
joneubank, line 1: only test file not in the tests directory. can we move it?

packages/data-provider/src/utils/submissionUtils.ts
```diff
@@ -386,35 +308,246 @@ export const groupSchemaErrorsByEntity = (input: {
...
+				...(conflictErrors.deletes[entityName] ?? []),
+				...deleteRecordIds.map((recordId) => ({ recordId, errors: [conflictErrorFor(systemId, 'UPDATE')] })),
+			];
+		});
+	});
+
+	return conflictErrors;
+};
+
+export type DeleteStagingConflicts = {
```
joneubank, line 392: type name is not descriptive of contents. This feels specific enough that maybe it doesn't need to be exported and can just be used within a single module?

packages/data-provider/src/utils/submissionUtils.ts
joneubank, line 1: This has become a 1000 line file with a very mixed focus. It would be best to organize these functions into files with a more defined scope. You could perhaps have a `/submissionUtils` path with many of these functions grouped into files based on their focuses.

packages/data-provider/src/utils/types.ts
joneubank, line 1: Same comment here as for the submission utils. This file is now massive and covers too many disparate uses. It would be better for many of these types to live closer to where they are defined (ex. repository types with their repositories or the same dir). Some of these are utility types (ex. `Keys<>`), and many of these are request and response types that should be defined with the controllers that actually need them.

packages/data-provider/test/integration/routers/dictionary/dictionaryMigration.spec.ts
```diff
@@ -14,6 +14,8 @@ import {
 	VALID_DICTIONARY_VERSION,
 } from './fixtures.js';
 
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
```
joneubank, line 17: I would leave a comment explaining why we are defining a sleep function. It made me nervous to see sleep function during a test, since these if used in unit tests can result in tests being fragile (fail if not slept long enough) or long running (tests take a long time cause much sleeping). This could be the case here, but there seems to be no other way to detect if a migration has actually completed.

Alternatively, jsut define this within the function where it is used so that its usage is immediately clear.

packages/data-provider/test/integration/routers/dictionary/dictionaryMigration.spec.ts
```diff
@@ -43,6 +45,23 @@ describe('Integration - Dictionary Migration', () => {
 	const registerDictionary = async (payload: RegisterPayload, force = false) =>
 		appDictionary.post(`/register${force ? '?force=true' : ''}`).send(payload);
 
+	// The migration itself runs in a background worker (fired without being awaited by
+	// `registerDictionary`), so a test that mutates a migration's status directly must first wait
+	// for that worker to reach a terminal status — otherwise the manual override races with the
+	// worker's own status transition.
```
joneubank, line 51: 👍

packages/data-provider/src/repository/submissionRecordsRepository.ts
```diff
@@ -0,0 +1,368 @@
...
+							const [updatedRecord] = await executor
+								.update(submissionRecords)
+								.set({ state: 'INVALID', errors: errors ?? null })
+								.where(eq(submissionRecords.id, id))
+								.returning({ id: submissionRecords.id });
+							return updatedRecord?.id;
+						}),
+					);
+					updatedIds.push(...invalidUpdates.filter((id): id is number => id !== undefined));
+				}
```
joneubank, line 282: This could be a very slow step if there are many records with errors. I don't know what to suggest as an alternative, doing a batch like the previous two updates would be preferable... Might be worth a todo comment or flagging this for performance testing. Functionally fine though.

packages/data-provider/src/repository/submissionRecordsRepository.ts
```diff
@@ -0,0 +1,368 @@
...
+					inArray(submissionRecords.fileId, fileIds),
+					filterOptions?.actionTypes ? inArray(submissionRecords.actionType, filterOptions.actionTypes) : undefined,
+					filterOptions?.states?.length ? inArray(submissionRecords.state, filterOptions.states) : undefined,
+				),
+			)
+			.orderBy(submissionRecords.id);
+
+		if (paginationOptions) {
+			query.limit(paginationOptions.pageSize).offset((paginationOptions.page - 1) * paginationOptions.pageSize);
+		}
```
joneubank, line 63: This may not be an issue, but, when pagination is applied there is no information returned about the total number of files available. Usually a paginated request has a paginated response object that includes the full request summary information.

joneubank (commented):
This is an incredible piece of work, many places are affected by this change and it looks like you've captured most of them.

A lot of my comments have to do with code organization and maintainability, very few functional changes found.

---

One question that I don't know where else to put:

What happens when a new file is received while a validation is in progress? Is a new validation kicked off after the first completes? Does it invalidate the ongoing validation process?

packages/data-provider/src/repository/submissionRecordsRepository.ts
```diff
@@ -0,0 +1,368 @@
...
+		deleteByIds: async (
+			ids: number[],
+			tx?: PgTransaction<PostgresJsQueryResultHKT, SubmissionRecord, ExtractTablesWithRelations<SubmissionRecord>>,
+		): Promise<number> => {
+			try {
+				return await (tx || db).delete(submissionRecords).where(inArray(submissionRecords.id, ids));
+			} catch (error) {
+				logger.error(LOG_MODULE, `Failed deleting Submission Record by ids '${ids}'`, error);
+				throw new ServiceUnavailable();
+			}
```
joneubank, line 344: Apparently if `submissionRecords.id` is an empty array, this will produce a SQL syntax error. In this case, ServiceUnavailable is the wrong error to throw. Could we add a guard against this case where no delete action is sent (since no ids to delete)? Could just `return 0` since `ids.length === 0`

packages/data-provider/src/services/submission/submissionProcessor.ts
```diff
@@ -677,32 +666,77 @@ const createSubmissionProcessor = (dependencies: BaseDependencies) => {
...
+				for (const entityName of entityNames) {
+					const savedFileId = await submissionFilesRepository.save(
+						{
+							entityName: entityName,
+							fileName: genericSubmissionFileName(),
+							fileSize: getSizeInBytes(JSON.stringify(recordsParsed)),
+							submissionId: submission.id,
+						},
+						tx,
+					);
```
joneubank, line 703: It is not clear to me why we are adding multiple submission file records here. This seems related to updates to ID fields on a record, which result in a deletion and an insertion, but its not clear to me how we get multiple entity names (since we are only working with edits for a single schema). Moreover, why would each of these files have the same file name and file size? There was only one copy of that file submitted and now it is being recorded for multiple entities?

Maybe the description comment for this function could be added to so it can describe the purpose of this funciton and the expected side effects (what files are registered in the database).

packages/data-provider/src/services/submission/submissionService.ts
```diff
@@ -253,17 +270,30 @@ const submissionService = (dependencies: BaseDependencies) => {
 			};
 		}
 
-		const totalRecords = await submissionRepository.getTotalSubmissionsByCategory(categoryId, filterOptions);
+		const totalSubmissions = await submissionRepository.getTotalSubmissionsByCategory(categoryId, filterOptions);
+		const result = await Promise.all(
+			recordsPaginated.map(async (response) => {
+				const submissionRecordsSummary = await submissionRecordsRepository.getRecordsSummaryBySubmissionId(response.id);
```
joneubank, line 276: Individual query per submission is a lot of requests to the database. Could these be batched to retrieve many records by an array of IDs?

packages/data-provider/src/services/submission/submissionProcessor.ts
```diff
@@ -795,101 +848,146 @@ const createSubmissionProcessor = (dependencies: BaseDependencies) => {
...
+				.map((record) => record.id);
+
+			// Update records with validation state, marking records with errors as 'INVALID' and records without errors as 'VALID'
+			await submissionRecordsRepository.updateValidationState(
+				{
+					invalidRecords,
+					validRecordIds: recordsWithoutError,
+				},
+				tx,
+			);
```
joneubank, line 886: Potentially data issue - is it possible that more records have been added to this submission since the validation started? Imagine a scenario where we start a validation and read 100 records, that processes asynchronously and additional files are added to the submission so now there are 125 records. Now, at line 874, we read in 125 records of which only 100 have had validation, so all the new 25 are marked valid.

Is this correct? If so, I think we need to review how we determine validity instead of reading from the DB and creating the compiment of the record with errors.

packages/data-provider/test/unit/utils/result.spec.ts
joneubank, line 1: This is probably over testing. No harm, but this is just testing if the TS interface is adhered to, which it must be to compile.

packages/data-provider/test/assertions.ts
```diff
@@ -0,0 +1,10 @@
+import { expect } from 'chai';
+
+/**
+ * Asserts that the value is not null or undefined.
+ * This lets TypeScript know the value is of type T after this check.
+ * @param value The value to check for existence.
+ */
+export function assertExists<T>(value: T | null | undefined): asserts value is T {
+	expect(value).to.exist;
+}
```
joneubank, line 10: This is a really cool tool, I will be reusing this in other places. I didn't know of assertion functions.

joneubank (changes requested):
Couple more notes from this morning.

---
Another note: there are no integration tests added for the new repositories or the modified controllers.

Some examples:
- checks that an uploaded file gets the correct number of records added to the submission records table
- after receiving a new file, a worker is dispatched for validation
- after validation completes, each record state is updated as expected
- when deleting a file, the database is changed correctly (submission file removed, correct number of submission records removed)

packages/data-provider/src/repository/submissionRecordsRepository.ts
leoraba: changed code as suggested for consistency.

packages/data-provider/src/repository/submissionRecordsRepository.ts
leoraba: updated the tsdocs to describe better the behaviour of this function

packages/data-provider/src/services/submission/submissionProcessor.ts
leoraba: changed this log to info.

packages/data-provider/src/services/submission/submissionService.ts
leoraba: Restructed the tsdoc text into clear sections including a short summary, a bullet list and return descriptions. 

Verified this is displayed correctly in vs codoe.

packages/data-provider/src/services/submission/submissionService.ts
leoraba: updated this tsdoc, to separate sections, will make sure to use this a standard format along the code.

packages/data-provider/src/services/submission/submissionService.ts
leoraba: Updated the function’s TSDoc using standard formatting and documented the conditions under which it throws `BadRequest` and `InternalServerError`

packages/data-provider/src/utils/submissionUtils.ts
leoraba: Renamed the type to `DeleteStagingConflictResolution` and removed the `export` since it's not being used externally.

packages/data-provider/src/services/submission/submissionProcessor.ts
leoraba: todo implemented, this code now returns the entityName as expected

packages/data-provider/src/services/submission/submissionService.ts
leoraba: changed this code to handle muliple IDs in a single query, using new function `getRecordsSummaryBySubmissionIds`

packages/data-provider/src/services/submission/submissionService.ts
leoraba: - the function `createSubmissionSummaryResponse` formats the entire single submission, in this code it's being called for each submission in the paginated response. The `create` prefix in the name is consistent with the application convention for preparing raw data for an API response format.
- Added explicitely the type `SubmissionSummaryResponse[]` as suggested to make the code clearer

packages/data-provider/src/repository/submissionRecordsRepository.ts
leoraba: renamed this type to `SubmissionRecordAggregate` and added a descriptive tsdoc of its meaning.

packages/data-provider/src/repository/submissionRecordsRepository.ts
leoraba: created the type `RepositoryTransaction` as suggested and placed in a common file so other repositories can use it as well.

packages/data-provider/src/repository/submissionRecordsRepository.ts
leoraba: updated to return the `id` of the deleted record instead. In case the caller want to know what ids were successfully deleted.

packages/data-provider/src/repository/submissionRecordsRepository.ts
leoraba: This endpoint is intentionally focused on returning only the requested submission records, while summary information is provided separately in another function. The summary includes aggregate counts, and this endpoint allows users to fetch the specific records they need. However, if the API’s pagination contract requires total counts or pagination metadata in every paginated response, this method should be updated to return that metadata as well.

packages/data-provider/src/repository/submissionRecordsRepository.ts
leoraba: good catch. code updated to return `0` in that case.

packages/data-provider/src/repository/submissionRecordsRepository.ts
leoraba: added a TODO to improve performance on this logic

packages/data-provider/src/repository/submissionRecordsRepository.ts
leoraba: will work on this and other performance TODOs in a separate PR

packages/data-provider/src/repository/submissionRecordsRepository.ts
leoraba: organized the functions as suggested, sorted alphabetically to find them easier.

packages/data-provider/src/utils/types.ts
leoraba: just moved out the types relevant to this PR, to avoid changes out of the scope.

packages/data-provider/src/utils/submissionUtils.ts
leoraba: moved out only the relevant code to this PR to a different file `submissionRecordUtils.ts`, this reduced the length from 867 to 595.

packages/data-provider/test/integration/routers/dictionary/dictionaryMigration.spec.ts
leoraba: `sleep` function is being used in 3 tests, to avoid duplicating code I moved to `test/integration/utils.ts` adding comments wherever is being used.

The purpose is to wait for async process like: Dictionary migration and Data validation which runs in a worker thread.

packages/data-provider/src/services/submission/submissionProcessor.ts
leoraba: This is a tricky but real scenario, added a comment on this code to add clarification, but basically, an edit of a single record could impact dependants records causing cascade effect (e.g. editing `participant` and changing its ID cascades an FK update into `specimen`) in which case a Delete and Insert is added, so editing one entity record can genuinely produce update records destined for a completely different entity table.

Furthermore, each file refers one single entity, this could mean a "synthetic" file is being created as a side effect per affected entity.

Regarding the `fileSize` value, it was a legitimate bug, the code was updated to include solely the records regarding that entity.

packages/data-provider/src/services/submission/submissionProcessor.ts
leoraba: Code updated to include to update the submission status to `VALIDATING` inside the worker function `dataValidation`, this makes sure the status is updated when it's actually inside the validation process

packages/data-provider/src/services/submission/submissionProcessor.ts
leoraba: code updated, the submission is set to `VALIDATING` automatically inside the function `dataValidation`

packages/data-provider/src/services/submission/submissionProcessor.ts
leoraba: fixed this bug, now it returns the value directly from the transaction

packages/data-provider/src/services/submission/submissionProcessor.ts
leoraba: code fixed!

packages/data-provider/src/services/submission/submissionProcessor.ts
leoraba: an edge case scenario but worth fix it. thanks for pointing this out, the code has been updated to receive a list of ids that were validated instead of querying the database again.

packages/data-provider/test/unit/utils/submission/submissionInsertDataFromFiles.spec.ts
leoraba: test moved to the corresponding test file

leoraba:
PR updated to address the feedback received . It's ready for a second review 🎉

PR #212, opened 2026-07-17
fix: surface file parse errors with fault isolation and structured logging

justincorrigible:
Each file in a batch submission is now processed independently: a stream error or schema validation failure on one file is caught and reported without blocking the remaining files.

Parse results use a three-state discriminator (ok/invalid/error) and are returned synchronously in the API response, so callers know immediately which files failed and why.

Structured logging separates human-readable console output (key=value) from a JSON file transport (LOG_JSON env var) for log aggregator ingestion. Field names and line numbers are logged for schema errors; field values are not (OWASP A03).

Bonus: Adds optional sync=false query parameter to the file upload endpoint to allow background parsing for very large batches; when omitted or true, behaviour is synchronous and unchanged from the existing default.

## Readiness Checklist

- [x] **Self Review**
- [x] **Formatting**
  - Automated code formatters (ie. Prettier) have been run
- [x] **Local Testing**
- [x] **Updated Tests**
- [x] Documentation

### DIFF
--- a/packages/data-provider/src/config/logger.ts
+++ b/packages/data-provider/src/config/logger.ts
@@ -65,7 +65,7 @@ export const getLogger = (config: LoggerConfig): Logger => {
 		}),
 	);
 
-	// Plain-text file transport (legacy; kept for backwards compatibility).
+	// Plain-text file transport for pre-Kubernetes deployments that relied on file-based log shipping.
 	if (config.file) {
 		transportList.push(new transports.File({ filename: 'logs.log' }));
 	}
--- a/packages/data-provider/src/config/logger.ts
+++ b/packages/data-provider/src/config/logger.ts
@@ -1,4 +1,4 @@
-import { createLogger, format, LoggerOptions, transports } from 'winston';
+import { createLogger, format, transports } from 'winston';
 
 import { LoggerConfig } from './config.js';
 
@@ -12,29 +12,58 @@ export type Logger = {
 };
 
 /**
- * Get a Logger instance for log messages
- * @param config logger configuration
- * @returns functions to log messages based on each log level
+ * Splits a variadic argument list into a message string and optional structured metadata object.
+ * When the last argument is a plain object (not an Error, array, or null), it is treated as metadata.
+ */
+const splitArgs = (messages: unknown[]): { message: string; meta?: Record<string, unknown> } => {
+	const last = messages[messages.length - 1];
+	const isMetaObject =
+		last !== null &&
+		last !== undefined &&
+		typeof last === 'object' &&
+		!Array.isArray(last) &&
+		!(last instanceof Error);
+	if (isMetaObject && messages.length > 1) {
+		return { message: messages.slice(0, -1).join(' - '), meta: last as Record<string, unknown> };
+	}
+	return { message: messages.join(' - ') };
+};
+
+/** Renders a metadata object as `key=value` pairs for human-readable console output. */
+const metaToString = (meta: Record<string, unknown>): string =>
+	Object.entries(meta)
+		.map(([k, v]) => `${k}=${typeof v === 'string' ? v : JSON.stringify(v)}`)
+		.join(' ');
+
+/**
+ * Get a Logger instance for log messages.
+ * Console transport emits human-readable `key=value` metadata.
+ * When `config.json` is true, a separate JSON file transport writes structured logs to `logs/app.json`
+ * for ingestion by log aggregators (e.g. Kibana).
  */
 export const getLogger = (config: LoggerConfig): Logger => {
-	const transportList: LoggerOptions['transports'] = [];
+	const { combine, timestamp, colorize, errors, json, printf } = format;
 
-	const { combine, timestamp, colorize, printf } = format;
-
-	// console transport
-	const consoleLog = new transports.Console({
+	const consoleTransport = new transports.Console({
 		format: combine(
+			errors({ stack: true }),
 			timestamp(),
 			colorize(),
-			printf(({ timestamp, level, message }) => `${timestamp} [${level}]: ${message}`),
+			printf(({ timestamp, level, message, meta }) => {
+				const metaPart = meta ? ` ${metaToString(meta as Record<string, unknown>)}` : '';
+				return `${timestamp} [${level}]: ${message}${metaPart}`;
+			}),
 		),
 	});
-	transportList.push(consoleLog);
 
-	// file transport
-	if (config.file) {
-		const fileLog = new transports.File({ filename: 'logs.log' });
-		transportList.push(fileLog);
+	const transportList = [consoleTransport];
+
+	if (config.json) {
+		const jsonTransport = new transports.File({
+			filename: 'logs/app.json',
+			format: combine(errors({ stack: true }), timestamp(), json()),
+		});
+		transportList.push(jsonTransport);
 	}
 
 	const logger = createLogger({
@@ -42,23 +71,17 @@ export const getLogger = (config: LoggerConfig): Logger => {
 		transports: transportList,
 	});
 
-	const log = (...message: unknown[]) => {
-		const fullMessage = message.join(' - ');
-		return fullMessage;
-	};
+	const emit =
+		(fn: (message: string, meta?: object) => void) =>
+		(...messages: unknown[]) => {
+			const { message, meta } = splitArgs(messages);
+			fn(message, meta ? { meta } : undefined);
+		};
 
 	return {
-		debug: (...messages: unknown[]) => {
-			return logger.debug(log(...messages));
-		},
-		warn: (...messages: unknown[]) => {
-			return logger.warn(log(...messages));
-		},
-		info: (...messages: unknown[]) => {
-			return logger.info(log(...messages));
-		},
-		error: (...messages: unknown[]) => {
-			return logger.error(log(...messages));
-		},
+		debug: emit((msg, meta) => logger.debug(msg, meta)),
+		warn: emit((msg, meta) => logger.warn(msg, meta)),
+		info: emit((msg, meta) => logger.info(msg, meta)),
+		error: emit((msg, meta) => logger.error(msg, meta)),
 	};
 };
--- a/packages/data-provider/src/config/logger.ts
+++ b/packages/data-provider/src/config/logger.ts
@@ -1,4 +1,4 @@
-import { createLogger, format, transports } from 'winston';
+import { createLogger, format, transport, transports } from 'winston';
 
 import { LoggerConfig } from './config.js';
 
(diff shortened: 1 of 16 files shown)


packages/data-provider/src/utils/submissionUtils.ts
```diff
@@ -831,32 +832,50 @@ export const segregateFieldChangeRecords = (
 	);
 };
 
+/** Per-file outcome from `submissionInsertDataFromFiles`. */
+export type FileParseResult =
+	| { status: 'ok'; fileName: string; entityName: string }
+	| { status: 'invalid'; fileName: string; entityName: string; parseErrors: ParseSchemaError[] }
+	| { status: 'error'; fileName: string; entityName: string; streamError: string };
```
joneubank: Since all cases have the same `fileName` and `entityName` properties , just to DRY this up a bit, it may be easier to read like:
```ts
export type FileParseResult = {
	fileName: string;
	entityName: string;
} & (
	| { status: 'ok' }
	| { status: 'invalid'; parseErrors: ParseSchemaError[] }
	| { status: 'error'; streamError: string }
);
```

Alternately, we can use the `Result` type that already exists in this repo is well setup to return a binary result (success/failure). TGo model a case with multiple failure states by setting the failure data type to be a discriminated union between schema validation errors and streaming errors:

```ts
export type FileParseResult = {
	fileName: string;
	entityName: string;
} & Result<void, { status: 'INVALID'; parseErrors: ParseSchemaError[] } | { status: 'STREAM_ERROR'; streamError: string }>;

// example definition using the repo's `success` and `failure` helpers:
const ok: FileParseResult = { fileName: 'donor.tsv', entityName: 'donor', ...success(null) };
const invalid: FileParseResult = { fileName: 'donor.tsv', entityName: 'donor', ...failure({status: 'INVALID', parseErrors}) };
const errors: FileParseResult = { fileName: 'donor.tsv', entityName: 'donor', ...failure({status: 'STREAM_ERROR', streamError }) };

// reading from a parseResult value that is a FileParseResult
const parseResult: FileParseResult = parseFile(someFile);

console.log(`We have the result for ${parseResult.fileName} with entity ${parseResult.entityName}`);


if(parseResult.success) {
	// result.data is void
console.log(`No schema errors found.`);
} else {
	switch(result.data.status) {
		case 'STREAM_ERROR': {
			console.log(`A streaming error occurred while parsing:`, result.data.streamError);
			break;
		}
		case 'INVALID': {
			console.log(`There were schema errors found when parsing the file:`, result.data.parseErrors);
			break;
		}
	}
}
```

There is a third option, which will update the Result type to natively support defining more than binary outcomes. This is the version of the Result type that we have defined and use in PCGL and OHCRN. It would require some minor type refactoring but would accomplish the same thing in a library standardized way. I would recommend this if you see many future uses of having 3 or more possible Result cases in functions in the code base (likely).

leoraba (approved):
LGTM

packages/data-provider/src/config/logger.ts
```diff
@@ -12,53 +12,90 @@ export type Logger = {
...
+					const metaKeys = Object.keys(meta).filter((k) => k !== 'stack');
+					const metaStr = metaKeys.length ? ' ' + metaToString(meta as Record<string, unknown>) : '';
+					return `${timestamp} [${level}]: ${message}${metaStr}`;
+				}),
+			),
+		}),
+	);
 
-	// file transport
+	// Plain-text file transport (legacy; kept for backwards compatibility).
```
joneubank: How is this a backwards compatibility concern? Its just a logger config option, no?

packages/data-provider/src/utils/submissionUtils.ts
justincorrigible: updated the PR to make it DRYer. that is legitimately actionable feedback right now.
On using the existing Result type, the change is structurally murkier than it looks:
The `Result` type is designed for binary success/failure as you correctly point out, and forcing a three-outcome case into it means the success arm carries data: void (i.e. data: undefined), which is awkward, and the failure arm is a second discriminated union nested inside data.
Callers switch from narrowing on `status === 'ok'` to `result.success === true`, and everything downstream needs updating.  There is that other approach with the switch example, which I like a lot and I think is a clever long-term direction, especially as the team wants to standardize `Result` (and logging in general) as the repo's multi-outcome type... but doing it here buries a structural decision inside a bug-fix PR.

Better as a separate, deliberate refactor.

packages/data-provider/src/config/logger.ts
```diff
@@ -12,53 +12,90 @@ export type Logger = {
...
+	}
+
+	// JSON file transport for log aggregators (Kibana via Filebeat, etc.).
+	if (config.json) {
+		transportList.push(
+			new transports.File({
+				filename: 'logs/app.json',
+				format: combine(timestamp(), errors({ stack: true }), json()),
+			}),
+		);
```
joneubank, line 80: We need to be careful with file transports for when these run inside kubernetes pods. I remember having issues where the pod crashes because it doesnt have any kind of file system write access, and it also won't have generated log files inside the pods being captured by the logging service. I think the correct approach when we configure the output as JSON is to replace the output of the console transport with this formatting, leaving the human readable version for local dev environments.

packages/data-provider/src/services/submission/submissionProcessor.ts
```diff
@@ -865,12 +890,18 @@ const createSubmissionProcessor = (dependencies: BaseDependencies) => {
 			// Perform Schema Data validation in a worker thread
 			dependencies.workerPool.dataValidation({ submissionId: activeSubmission.id });
 		} catch (error) {
-			logger.error(`There was an error processing submitted files: ${fileSummaries}`, JSON.stringify(error));
+			logger.error(LOG_MODULE, `Error processing submitted files`, {
+				files: fileSummaries,
+				error: error instanceof Error ? error.message : String(error),
+				errorType: error instanceof Error ? error.name : 'unknown',
+			});
```
joneubank, line 897: Catch and log, no handling.

When we throw a BadRequest due to no active submission found, we are already done parsing the file and have included those results in the fileResults object. In that case, this function returns that it completed successfully but actually stopped processing after `submissionInsertDataFromFiles` but before `mergeInsertsRecords` and `submissionRepository.update`. Likely this will communicate the wrong message to its caller?

At a glance, maybe we want to perform the check for an active submission at the top of the function and return out quickly before doing any work if no active submission is found.

packages/data-provider/src/services/submission/submissionService.ts
```diff
@@ -503,12 +504,14 @@ const submissionService = (dependencies: BaseDependencies) => {
 		organization,
 		username,
 		fileEntityMap,
+		sync = false,
```
joneubank, line 503: The ts docs for this function declare each parameter. We just added a parameter so we need to update the docs.

My recommendation: do not include parameters in TSDocs unless they are ambiguous and require a description, that way changes to the function signature don't fall out of line with the documentation. You can always add notes just for the params that are confusing. TS provides the type information for our params so TSDocs aren't needed for this.

packages/data-provider/src/utils/fileUtils.ts
```diff
@@ -170,51 +149,55 @@ type FileProcessingResult = {
...
- * @returns A `Promise<FileProcessingResult>` that resolves to an object containing two arrays:
- * - `validFiles`: Files that have a `.tsv` extension and contain the `systemId` header.
- * - `fileErrors`: Files that either have an invalid extension or are missing the required `systemId` header.
- */
-export async function processFiles(files: Express.Multer.File[]): Promise<FileProcessingResult> {
-	const result: FileProcessingResult = {
-		validFiles: [],
-		fileErrors: [],
-	};
+type FileOutcome = { valid: true; file: Express.Multer.File } | { valid: false; error: BatchError };
```
joneubank: We have the `Result` type for exactly this situation:

```ts
import { failure, success, type Result } from './result.js';

type FileOutcome = Result<Express.Multer.File, BatchError>;

const classifyFile = async (file: Express.Multer.File): Promise<FileOutcome> => {
	try {
		if (!getSubmittedFileType(file).success) {
			return failure({
				type: BATCH_ERROR_TYPE.INVALID_FILE_EXTENSION,
				message: `File '${file.originalname}' has invalid file extension. File extension must be '${SUPPORTED_FILE_EXTENSIONS.options}'`,
				batchName: file.originalname,
			});
		}

		const headers = await readHeaders(file);
		return headers.includes('systemId')
			? success(file)
			: failure({
					type: BATCH_ERROR_TYPE.MISSING_REQUIRED_HEADER,
					message: `File '${file.originalname}' is missing the column 'systemId'`,
					batchName: file.originalname,
				});
	} catch {
		return failure({
			type: BATCH_ERROR_TYPE.FILE_READ_ERROR,
			message: `Error reading file '${file.originalname}'`,
			batchName: file.originalname,
		});
	}
};
```

and then:
```ts
export const processFiles = async (files: Express.Multer.File[]): Promise<FileProcessingResult> => {
	const outcomes = await Promise.all(files.map(classifyFile));
	return {
		validFiles: outcomes.filter((o) => o.success).map((o) => o.data),
		fileErrors: outcomes.filter((o) => !o.success).map((o) => o.data),
	};
};
```

Note that you do not need to define the type predicate (`is SomeType`) in an array.filter function if the boolean you return successfully narrows the type.

packages/data-provider/src/utils/schemas.ts
```diff
@@ -1,6 +1,6 @@
 import type { ParamsDictionary } from 'express-serve-static-core';
 import type { ParsedQs } from 'qs';
-import { z } from 'zod';
+import { z as zod } from 'zod';
```
joneubank, line 3: 👍 nice, thanks.

joneubank (changes requested):
One critical change is that I think the logger's JSON file output could have issues with many deploy scenarios, consider just changing the console transport formatting?

One likely important change to submissionProcessing to prevent us from doing partial work then returning as though we have succeeded.

Last two comments are just TS stuff ;)

packages/data-provider/src/config/logger.ts
justincorrigible: support for older systems (pre k8s)

packages/data-provider/src/config/logger.ts
justincorrigible: that's completely right! something to account for when we "enable" the feature. right now we're just adding support for it, not enabled by default

that said, "replacing the output" is not the right approach, in that the output should be plainly readable by humans, hence the transport separation

packages/data-provider/src/utils/fileUtils.ts
justincorrigible: good point! ~this is pre-existing code, though, that is beyond the scope of this PR: a bug fix.
adding it to the tech-debt doc as something to improve later~ 👍 

further inspection, I was wrong, and this can (or should?) definitely be managed now. let me know what you think

packages/data-provider/src/services/submission/submissionProcessor.ts
justincorrigible: nice! you're right. switching the order of operations here to account for that

packages/data-provider/src/services/submission/submissionService.ts
justincorrigible: not entirely clear what you're asking here, but it sounds like I'm meant to switch from TSDocs to verbose comment, so that's what I'll do

justincorrigible:
checks "failed" due to Jenkins Agent Nodes issues (timeout, possibly related to a 0B swap space sign in the Built-in one 🤷)... unsure how to fix those from this end. may need @yalturmes's help on this one

packages/data-provider/src/services/submission/submissionProcessor.ts
joneubank: This works. The pattern that enables this error is having the return object declared outside of the scope of the try/catch block, that allows it to be modified AND returned without respecting the errors thrown.

joneubank (approved):
Thanks for the updates. Good to go!

PR #213, opened 2026-07-22
Category alias, plus routing Kafka messages by categoryId when a topic is shared

justincorrigible:
Adds category alias as a first-class feature (creation-time, resolvable via GET, assign/unassign after the fact), and publishes `categoryId`/`categoryAlias` on every committed Kafka message so a consumer sharing one topic across multiple categories, like Maestro, can route correctly (solving the wrong-index bug this whole thing started from).

The commits are organized in deliberate layers, each with its own description, rather than as one flat diff, to make this easier to review incrementally instead of all at once:

1. Agentics `devctx` upgrade (unrelated tooling bump, picked up in the same session)
2. Schema: the `alias` column itself
3. Kafka messages carry `categoryId`/`categoryAlias` (the actual bug fix)
4. Category alias registrable at creation, readable via GET
5. `categoryId-or-alias` resolution wired into every other router
6. Assign/unassign endpoints for a category created before this feature existed
7. `.dev/` housekeeping (roadmap, tech debt, session log)

Maestro-side routing changes are in a [separate PR](https://github.com/overture-stack/maestro/pull/295) against that repo.


packages/data-model/src/models/dictionary_categories.ts
```diff
@@ -8,6 +8,7 @@ import { submittedData } from './submitted_data.js';
 export const dictionaryCategories = pgTable('dictionary_categories', {
 	id: serial('id').primaryKey(),
 	activeDictionaryId: integer('active_dictionary_id').notNull(),
+	alias: varchar('alias').unique(),
```
justincorrigible, line 11: nullable and unique: nullable because every category before this feature has no alias yet, unique so `getCategoryByIdOrAlias`'s alias lookup can never match more than one category.

packages/data-provider/src/utils/types.ts
```diff
@@ -383,14 +395,16 @@ export type SubmittedDataResponse = {
  * Result type Post-Commit Submission
  */
 export type ResultOnCommit = {
-	submissionId: number;
-	organization: string;
+	/** Not present if the category has no alias assigned. */
+	categoryAlias?: string;
```
justincorrigible, line 399: sourced from the submission's own category (`submission.dictionaryCategory`), not computed per commit: a commit always finalizes exactly one submission, and a submission is scoped to a single category throughout, so there's nothing to look up per record.

packages/data-provider/src/repository/activeSubmissionRepository.ts
```diff
@@ -48,12 +48,25 @@ const activeSubmissionRepository = (dependencies: BaseDependencies) => {
...
 				name: true,
 			},
 		},
 	} as const satisfies Record<string, { columns: BooleanTrueObject }>;
 
+	/**
+	 * Normalizes a queried record's `dictionaryCategory.alias` from the DB's `string | null`
+	 * to the public `CategorySummary` contract's `string | undefined` (omitted, not null, when unset).
+	 */
+	const withAliasNormalized = <T extends { dictionaryCategory: { alias: string | null } }>(record: T) =>
```
justincorrigible, line 62: the DB gives back `alias: string | null`, but the public `CategorySummary` contract wants `alias?: string` (omitted, not `null`, when unset). this normalizes that at the 4 call sites returning submission-shaped records.

packages/data-provider/src/repository/categoryRepository.ts
```diff
@@ -87,6 +114,43 @@ const repository = (dependencies: BaseDependencies) => {
 			}
 		},
 
+		/**
+		 * Find a Category by its numeric id or its alias; alias wins if both could match
```
justincorrigible: alias wins if a numeric id and a different category's alias could both match the same value. only actually happens if someone's alias happens to look like a valid id, allowed on purpose, an alias can be a data version label (e.g. `2026`), not just an id stand-in. the logic in this PR accounts for those scenarios... read on

packages/data-provider/src/services/dictionaryService.ts
```diff
@@ -133,6 +135,14 @@ const dictionaryService = (dependencies: BaseDependencies) => {
 		// Check if Category exist
 		const foundCategory = await categoryRepo.getCategoryByName(categoryName);
 
+		if (foundCategory && alias) {
+			// Alias assignment is creation-time only; reassigning an existing category's alias
+			// is a separate, later feature.
+			throw new BadRequest(
+				`Category '${categoryName}' already exists; alias can only be set when creating a new category`,
```
justincorrigible, line 142: alias assignment is creation-time only for now. this rejects supplying one while registering a new dictionary version onto a category that already exists, use `PUT /category/:categoryId/alias` afterward instead.

packages/data-provider/src/controllers/dictionaryController.ts
```diff
@@ -18,6 +18,7 @@ const controller = (dependencies: BaseDependencies) => {
 	return {
 		registerDictionary: validateRequest(dictionaryRegisterRequestSchema, async (req, res, next) => {
 			try {
+				const alias = req.body.alias;
```
justincorrigible, line 22: found this one the hard way (right before pushing): `alias` was never actually read from the request body here before, the whole register-with-an-alias feature had never worked end to end despite the service-layer logic being correct. only caught it once the integration suite ran against real Postgres.

ps: also added `alias` to the response, wasn't returned either, which is exactly why the original test for this endpoint didn't catch it.

packages/data-provider/src/repository/categoryRepository.ts
```diff
@@ -1,11 +1,17 @@
...
 import { Category, Dictionary, dictionaryCategories, NewCategory } from '@overture-stack/lyric-data-model/models';
 
 import { BaseDependencies } from '../config/config.js';
-import { ServiceUnavailable } from '../utils/errors.js';
+import { ServiceUnavailable, StatusConflict } from '../utils/errors.js';
+import { isValidIdNumber } from '../utils/formatUtils.js';
+
+// Postgres' unique_violation code, via the `pg` driver's `.code` field. A race between an
+// app-level uniqueness check and the write can still surface here, handled as 409, not 503.
+const isUniqueConstraintViolation = (error: unknown): boolean => {
```
justincorrigible, line 12: `save()`/`update()` used to map every DB error, including a unique-constraint race, to a generic 503. this detects Postgres' `23505` specifically and surfaces it as 409 instead, since the application-level pre-check can still lose a race under concurrent requests. this was a pre-existing issue, that needed a fix for my tests to pass

packages/data-provider/src/utils/resolveCategoryId.ts
```diff
@@ -0,0 +1,14 @@
+import { BaseDependencies } from '../config/config.js';
+import categoryRepository from '../repository/categoryRepository.js';
+
+/**
+ * Resolves a categoryId path param (numeric id or alias) to the category's numeric id; alias
+ * wins on a collision (see `getCategoryByIdOrAlias`). Returns undefined if neither matches.
+ */
+export const resolveCategoryId = async (
```
justincorrigible, line 10: pulled this into its own file since 7 different controllers now need the same categoryId-or-alias resolution, cheaper than routing everything through `categoryService.getDetails`, which does an extra organizations lookup nothing here needs.

packages/data-provider/src/config/config.ts
```diff
@@ -42,7 +42,9 @@ export type IdServiceConfig = {
 };
 
 export type ValidatorEntry = {
-	categoryId: number;
+	/** A category's numeric id or its alias; matched against the request's categoryId by string
+	 * equality (see `findValidatorEntry`), not by shape. */
+	categoryId: number | string;
```
justincorrigible, line 47: widened so `VALIDATOR_CONFIG` can reference a category by alias too. matching stays representation-sensitive on purpose: config and request have to use the same form (both numeric, or both the same alias), same design as Maestro's own categoryId/alias matching.

packages/data-provider/src/services/categoryService.ts
```diff
@@ -3,42 +3,111 @@ import * as _ from 'lodash-es';
...
+		 * requesting the alias it already has succeeds as a no-op, not a conflict. Changing an
+		 * existing alias to a new value isn't supported yet, see `.dev/roadmap.md`.
+		 */
+		assignAlias: async (categoryIdOrAlias: string, alias: string, username?: string): Promise<CategorySummary> => {
+			const category = await categoryRepo.getCategoryByIdOrAlias(categoryIdOrAlias);
+			if (!category) {
+				throw new NotFound(`Category '${categoryIdOrAlias}' not found`);
+			}
+
+			if (category.alias === alias) {
```
justincorrigible, line 63: this is the idempotency case: a retried PUT with the same alias succeeds as a no-op rather than conflicting. without it, a client that doesn't know whether its first request landed would get a false 400 on retry.

packages/data-provider/src/utils/formatUtils.ts
```diff
@@ -45,6 +45,17 @@ export function isValidIdNumber(value: unknown): boolean {
...
 }
 
+/**
+ * Checks if a string is safe as a category alias: a non-empty URL-safe slug (letters,
+ * numbers, hyphens, underscores). Purely numeric values are allowed, an alias may be a
+ * data version label rather than an id, and is matched by equality, not shape.
+ * @param {string} value
+ * @returns {boolean}
+ */
+export function isValidCategoryAlias(value: string): boolean {
```
justincorrigible, line 58: purely numeric values are allowed on purpose, an alias might be a meaningful data version label, not just a stand-in for an id. matched by equality downstream, not by shape, so it's never confused with a real numeric id even when it looks like one.

.dev/roadmap.md
```diff
@@ -18,19 +18,43 @@ Add a `kafka` field to the `/health` response: `connected | disconnected | not c
...
 
 ### Kafka: message format design [pending discussion]
-Current message shape (`action`, `data`, `entityName`, `isValid`, `organization`, `systemId`) was designed around Maestro's expectations. Leo raised in PR #208 whether it is too Maestro-specific and should include additional fields (submission date, user, etc.) for other consumers. Jon was tagged but the thread is unresolved. Agree on a stable, consumer-agnostic schema before the format is treated as a public contract.
+Current message shape (`action`, `categoryAlias`, `categoryId`, `data`, `entityName`, `isValid`, `organization`, `systemId`) was designed around Maestro's needs. Open question from PR #208: too Maestro-specific, or should it include more fields (submission date, user) for other consumers? No backward-compatibility risk either way, Lyric never published to this topic before this feature existed; this is a design-quality question, not a break-something-today one.
+
+### Category alias: rename and transfer
```
justincorrigible, line 23: splitting this out because "reassign" as originally discussed was actually two operations with very different risk profiles bundled under one name. rename only touches one category; transfer moves an identity between two, a much bigger design question (grace period, audit trail, in-flight messages). didn't want the harder one to block shipping the easier one in this PR.

.dev/tech-debt.md
```diff
@@ -26,6 +39,18 @@ context: `server.ts` creates a logger with `getLogger()` and `buildAppConfig()`
 standalone: yes
 context: `FROM node:22-alpine` (Dockerfile line 9) is flagged with a high CVE by the IDE linter. Pre-existing, unrelated to Kafka work. Resolve by pinning to a patched digest or upgrading to a newer `node:22-alpine` release once available.
 
+### Category-level admin actions have no privilege check beyond "authenticated"
```
justincorrigible, line 42: honest gap, not something I'm fixing here: this and `dictionaryService.register()` change category state with no privilege check beyond "is this request authenticated". the codebase already has `isAdmin`/org-scoped helpers, just never wired up for category-level actions. logged rather than patched, since fixing only the new endpoints would leave `register()` inconsistent.

apps/server/swagger/categories-api.yml
```diff
@@ -24,6 +24,10 @@
 /category/{categoryId}:
   get:
     summary: Retrieve details of a specific category
+    description: >-
+      Accepts either the category's numeric id or its alias. If a numeric id could also match a
+      different category's alias, the alias wins, allowed on purpose since an alias may be a data
+      version label, not just a stand-in for an id.
```
joneubank: > If a numeric id could also match a different category's alias, the alias wins

This can't be right. If an alias can exactly match a numeric category ID then we could completely prevent being able to interact with that category ID. Do we not restrict aliases to never be integer values (valid category ID values)?

apps/server/swagger/dictionary-api.yml
```diff
@@ -19,18 +19,29 @@
...
             type: object
             properties:
+              alias:
+                type: string
+                description: >-
+                  An optional, stable, human-readable identifier for the category, usable anywhere
+                  a categoryId is accepted, in place of the numeric id. Letters, numbers, hyphens,
+                  and underscores only; an empty string is treated as not provided. A purely
+                  numeric alias is allowed, it may be a data version label rather than an id,
+                  matched by equality not shape. Can only be set here at category creation;
```
joneubank: > A purely numeric alias is allowed, it may be a data version label rather than an id, matched by equality not shape.

I thought we discussed not allowing integer aliases that would conflict with a category ID. This could open up some problematic or confusing behaviour if we allow this.

apps/server/swagger/schemas.yml
```diff
@@ -283,6 +288,19 @@ components:
...
+        alias:
+          type: string
+          description: Optional stable alias for the category. Not set if none is assigned.
+        id:
+          type: number
+          description: ID of the Category
+        name:
+          type: string
+          description: Name of the Category
+
```
joneubank, line 303: 👍

apps/server/swagger/schemas.yml
```diff
@@ -555,6 +566,9 @@ components:
     RegisterDictionaryResult:
       type: object
       properties:
+        alias:
+          type: string
+          description: Optional stable alias for the category. Not present if none was assigned.
```
joneubank: What makes this "stable"? Aliases can change at any time.

docs/02-lyric.md
```diff
@@ -9,7 +9,7 @@ Lyric is a tabular data management service designed to handle structured clinica
...
 - **Detailed change history:** Maintains a complete audit trail of all data modifications, tracking changes from committed submissions and updates ensuring data governance and accountability.
 - **SQON Query Endpoint:** Provides an endpoint for SQON (Structured Query Object Notation) based queries allowing complex search operations through combinations of simple field operations (`in`, `<=`, `>=`) and logic (`and`, `or`, `not`). This allows complex queries to be expressed in a simple JSON format.
-- **Multi-dictionary support:** Handles multiple Lectern dictionaries simultaneously, allowing organizations to manage different data categories with distinct schemas while maintaining data integrity and relationships
+- **Multi-dictionary support:** Handles multiple Lectern dictionaries simultaneously, allowing organizations to manage different data categories with distinct schemas while maintaining data integrity and relationships. A category can optionally have a stable alias, a human-readable identifier usable anywhere a numeric category ID is accepted
```
joneubank: "Stable" again, I'm not sure of the meaning here.

packages/data-provider/docs/kafka-publishing.md
```diff
@@ -23,6 +27,8 @@ Each message value is a JSON string:
 }
 ```
 
+`categoryAlias` is omitted entirely (not sent as `null`) when the category has no alias assigned.
```
joneubank, line 30: 👍

packages/data-provider/src/utils/resolveCategoryId.ts
```diff
@@ -0,0 +1,14 @@
+import { BaseDependencies } from '../config/config.js';
+import categoryRepository from '../repository/categoryRepository.js';
+
+/**
+ * Resolves a categoryId path param (numeric id or alias) to the category's numeric id; alias
+ * wins on a collision (see `getCategoryByIdOrAlias`). Returns undefined if neither matches.
```
joneubank:
```suggestion
 * Finds the category with an alias or ID matching the provided input string. This is commonly used 
 * to resolve the categoryId path param of many routes, where the value may be either contain a category ID or alias. 
 * If the input matches both a category ID and an alias, the alias has priority and will be returned. Returns undefined if no match can be found.
```

joneubank: - We should prioritize matches with Category ID. Category IDs are immutable and are therefore the most reliable way to access data in a category, these should never be able to be overwritten by an alias.
- This relies on a DB read but we are not handling any DB connection failures here. There is no way to distinguish between "system error" or "not found". This likely means an error is being thrown within the categoryService function that we call. If this is likely to throw an error, we should highlight this fact in the TS Docs to help avoid usage of this function outside of a try/catch block.

packages/data-provider/src/utils/schemas.ts
```diff
@@ -24,14 +24,18 @@ const booleanSchema = zod
...
-	.refine((value) => {
-		const parsed = parseInt(value);
-		return isValidIdNumber(parsed);
-	}, 'invalid category ID');
+	.refine((value) => isValidIdNumber(parseInt(value)) || isValidCategoryAlias(value), 'invalid category ID');
+
+const categoryAliasSchema = zod
+	.string()
+	.trim()
+	.refine((value) => value === '' || isValidCategoryAlias(value), 'alias must contain only letters, numbers, hyphens, and underscores');
```
joneubank: I would like to propose the following schemas to capture the types we are using clearly. What this does differently is:
- categoryId is treated as a number, which it is
- categoryAlias lets zod handle the regex check in the schema definition so that we don't need to maintain "isValidCategoryAlias" separately from this schema. This schema can be used to test if a value is valid category alias.
- creates a separate schema for handling the path parameter values that is a string which is valid for both the above types. it does this by defining a local scope category ID as string schema that relies on the categoryIdSchema to ensure the string, when converted to a number, is a valid category ID.

All this together creates a clean separation of responsibilities for validating the types of each value:

```ts
// Numeric Category ID
const categoryIdSchema = zod.number().int();

// Category Alias
const categoryAliasSchema = zod
	.string()
	.trim()
	.regex(/^[A-Za-z0-9_-]+$/, {
		message: 'Category alias must contain only letters, numbers, hyphens, and underscores.',
	});

// Schema that will confirm that a string contains a valid Category ID when parsed as a number
const categoryIdStringSchema = zod
	.string()
	.trim()
	.refine((value) => categoryIdSchema.safeParse(Number(value)).success, 'value not valid as a numeric Category ID');

// Category ID Path Parameter Schema - validates that a string value represents a valid category ID or alias
export const categoryIdOrAliasPathParamSchema = zod.union([categoryIdStringSchema, categoryAliasSchema]);
```

With this change you'll need to replace references of the `categoryIdSchema` in the route parameter schemas with the `categoryIdOrAliasPathParamSchema`

packages/data-provider/src/repository/categoryRepository.ts
joneubank: Repeating opinion from a previous comment:
I think that ID should take priority over Alias. IDs are immutable, they should be the most reliable way to identify and access data, it would be very difficult to deal with a scenario where a category alias changes what an ID actually points to. In fact, there may be no way to interact with such a category if it didn't already have a different alias.

packages/data-provider/src/repository/categoryRepository.ts
```diff
@@ -158,6 +222,13 @@ const repository = (dependencies: BaseDependencies) => {
 				}
 				return updated[0];
 			} catch (error) {
+				if (isUniqueConstraintViolation(error)) {
+					logger.warn(
+						LOG_MODULE,
+						`Update to Category '${categoryId}' conflicts with an existing name or alias (race with a concurrent request)`,
```
joneubank: RE: the parenthesis at the end of this message, how do we know this is a race condition?

packages/data-provider/src/services/categoryService.ts
```diff
@@ -3,42 +3,111 @@ import * as _ from 'lodash-es';
...
 			logger.debug(LOG_MODULE, `List all categories`);
-			return await categoryRepo.getAllCategoryNames();
+			const categories = await categoryRepo.getAllCategoryNames();
+			return categories.map(toCategorySummary);
+		},
+		/**
+		 * Sets a category's alias, only if it doesn't already have a different one. Idempotent:
+		 * requesting the alias it already has succeeds as a no-op, not a conflict. Changing an
+		 * existing alias to a new value isn't supported yet, see `.dev/roadmap.md`.
+		 */
```
joneubank, line 56: > only if it doesn't already have a different one

Does this mean that you can't set an alias on a category by using its current category alias as the identifier in the request? Could this be restated as "To set an alias you must use provide the category ID of the category that will be updated."

I think this is a good decision since there is an emphasis on idempotent behaviour, as this removes the non-idempotent behaviour of when we receive multiple requests to change from alias A to alias B (first request succeeds, second request is in error).

joneubank (changes requested):
Most important piece of feedback here is that we should prioritize matching the category ID or alias parameters to a category ID, instead of first matching to alias. This is to prevent the scenario where an alias matches an ID and prevents the use of that immutable category ID from referencing its category. IDs are the guaranteed reference to a specific piece of data and we should not allow using that identifier to point to something else.

apps/server/swagger/schemas.yml
justincorrigible: "does not change on its own" like an autogenerated value... rephrasing in a new commit

packages/data-provider/src/repository/categoryRepository.ts
justincorrigible: you're right, and I should've caught this myself, walking through it properly for others reading this in the future: 
if category A has id 5 with no alias, and category B later gets alias "5", /category/5 would've resolved to B instead of A. not just unreachable, silently misrouted to the wrong category's data. 

fixed now: `getCategoryByIdOrAlias` now checks numeric id first, `alias` operates only as fallback.
ids are permanent, aliases can be assigned/cleared/reassigned after the fact (this same PR adds that), so alias can never be allowed to win... rewrote also the collision test to assert the corrected direction, and ran it against actual Postgres to confirm using the integration tests

packages/data-provider/src/repository/categoryRepository.ts
justincorrigible: fair, we can't actually tell from a bare unique-constraint violation whether it was a genuine race or something else entirely. changed it so that both log lines now just say what happened (unique constraint violated) without asserting why

docs/02-lyric.md
justincorrigible: yeah, poor word choice on my part, I meant "doesn't change automatically the way an auto-generated token might," not "permanent." given assign/unassign exists in this same PR, "stable" actively overstates it

packages/data-provider/src/utils/schemas.ts
justincorrigible: this is a genuinely better shape than what's there now, splitting the numeric/alias/path-param concerns into three schemas instead of one confusingly-named `categoryIdSchema` that's really "id-or-alias as a string." retiring `isValidCategoryAlias` in favor of the native zod regex is the right call too, one less thing to keep in sync.

it's real added scope though, because `categoryIdSchema` gets referenced across every route that takes a `categoryId` path param, so this touches more than just the definition. want this in this PR or as its own focused follow-up? happy to do either, just didn't want to fold a schema refactor and all the side effects into an already significantly large PR without double checking first

packages/data-provider/src/services/categoryService.ts
justincorrigible: yes, exactly right, and your restatement is basically it: for the actual use case this feature exists for (a category that's never had an alias), you have to address it by id, since there's nothing else to identify it by yet. and you're right on the idempotency reasoning too, that's precisely why any existing alias blocks reassignment regardless of how you addressed the category to get there, otherwise two racing "change alias A to B" requests would have one succeed and one 400 on retry, instead of both safely resolving to the same end state

apps/server/swagger/dictionary-api.yml
justincorrigible: yes, we did. and I designed to allow integers regardless, thinking someone could want to use those as versions, or even to shadow a category on purpose... however the opposite scenario where you actually seek a category but someone else shadowed it by mistake is a serious consideration I missed. changing the code to prevent integers as aliases (even if presented as strings i.e. no 5 nor "5" allowed as alias)

apps/server/swagger/categories-api.yml
justincorrigible: You're right, and that doc line was wrong regardless of the alias-value question. Flipped the precedence so id wins on any collision, and separately pushed another fix that rejects purely numeric aliases outright at assignment time, so this collision shouldn't arise from normal use going forward either. A value containing a decimal point (e.g. 2.5) is still allowed as an alias, since it can never be mistaken for a whole-number id

packages/data-provider/src/utils/resolveCategoryId.ts
justincorrigible: kept your suggested wording but flipped which one wins to match. TSDoc now documents the `ServiceUnavailable` throw explicitly, so callers know not to treat `undefined` as the only failure mode

packages/data-model/migrations/0014_add_dictionary_category_alias.sql
leoraba, line 1: rename this file to a meaningful name, also update the `_journal.json` file accordingly.

packages/data-provider/src/controllers/dictionaryController.ts
leoraba, line 1: Since categoryId param could be an alias string or a number, we need to change all the controllers to accept both string and number, notice in this file there are 2 controllers that needs to change the the param category 
from: `const categoryId = Number(req.params.categoryId);`
to `const categoryIdOrAlias = req.params.categoryId;`

Also, similiar happens in `submissionController.ts` where some controllers converts `Number(req.params.categoryId)`

packages/data-provider/src/utils/schemas.ts
joneubank: i think its correct to make this update here, apply to all params where the meaning changes from "category ID" into "category ID or Alias". For the routes that still use only the category ID now, this will have a break as well since the parmas must be a string, for those I would use `categoryIdStringSchema` to keep previous functionality (I think this works).

packages/data-provider/src/services/categoryService.ts
joneubank: Great, thanks. I think this needs to be rephrased for clarity then "Sets a category alias. The category must be identified by ID. Category Alias can be changed at any time. To remove the alias, use the Delete Alias route." and then the Idempotent content.

packages/data-provider/src/controllers/dictionaryController.ts
joneubank: This is actually handled by ensuring the parameter is a string, but validating if the string is parseable as a number that its a valid int. This produces a string usable by the new "getCategoryByIdOrAlias" function in categoryService.

joneubank (commented):
Thanks for addressing these concerns. Left a couple comments, will come back after lunch to review and approve.

justincorrigible:
All very welcome feedback, thanks again guys... addressing in the next commit

packages/data-provider/src/controllers/dictionaryController.ts
justincorrigible: oh wow... trying to address this, Leo's finding goes broader than the two spots he flagged: 9 call sites total bypass alias resolution via `Number(req.params.categoryId)`, 2 in `dictionaryController.ts` and 7 in `submissionController.ts`: the string validation is correct, but it only confirms the value is shaped like a valid id or alias, it doesn't resolve it, and these 9 call sites never called `getCategoryByIdOrAlias` at all, so a valid alias silently became `NaN` downstream... anyway! fixed all of them to match the established `resolveCategoryId` pattern already used by `categoryController.ts`/`submittedDataController.ts`/`auditController.ts`.

while tracing it we ran across a related inconsistency: `categoryController.ts`'s own "not-found" case threw `BadRequest` (400) instead of `NotFound` (404), unlike every other controller handling the same "category doesn't exist" scenario. fixed that too, one line. 👍 BONUS

net effect: a nonexistent category id or alias now consistently returns 404 everywhere, instead of a mix of 400 and 404 depending which endpoint you hit. updated 3 existing tests that asserted the old, inconsistent status codes. full suite green: 273 unit, 99 integration.

packages/data-provider/src/services/categoryService.ts
```diff
@@ -3,42 +3,111 @@ import * as _ from 'lodash-es';
...
+			if (category.alias === alias) {
+				logger.info(LOG_MODULE, `Category alias unchanged`, { categoryId: category.id, alias, username });
+				return toCategorySummary(category);
+			}
+
+			if (category.alias) {
+				throw new BadRequest(
+					`Category '${category.name}' already has a different alias; changing an existing alias is not yet supported`,
+				);
+			}
```
joneubank, line 72: We can't modify aliases? This does match your comment, I now realize, but I'm surprised this is the decision. We can't modify an alias once set, so hope you don't make any typos :P

packages/data-provider/src/services/categoryService.ts
```diff
@@ -3,42 +3,111 @@ import * as _ from 'lodash-es';
...
-			return await categoryRepo.getAllCategoryNames();
+			const categories = await categoryRepo.getAllCategoryNames();
+			return categories.map(toCategorySummary);
+		},
+		/**
+		 * Sets a category's alias, only if it doesn't already have a different one. Idempotent:
+		 * requesting the alias it already has succeeds as a no-op, not a conflict. Changing an
+		 * existing alias to a new value isn't supported yet, see `.dev/roadmap.md`.
+		 */
+		assignAlias: async (categoryIdOrAlias: string, alias: string, username?: string): Promise<CategorySummary> => {
```
joneubank, line 57: If updating this to be "must assign alias using a category iD, then we should modify the function signature to have parameter `(categoryId: number` instead of the `categoryIdOrAlias`

joneubank (approved):
Can approve at this state. Comments for behaviour of assigning alias but it is working as is.

packages/data-provider/src/services/categoryService.ts
justincorrigible: it's perhaps unclear as state in the error...
as of right now, you can add and remove aliases, which enables reassigning them in a clunky way.
roadmap tracks an actual (future) idea to automate the "move this alias from A to B" that was beyond the current work requirements, strictly speaking

edit: that is to say this will be changed at that point, because it'll be made irrelavant

packages/data-provider/src/services/categoryService.ts
justincorrigible: TL;DR: narrowing to `categoryId: number` would mean resolving the `alias-or-id` earlier, in the controller, the same pattern `resolveCategoryId` already uses elsewhere, a real option if you'd like this consistent with the other controllers, but it's a relocation of the resolution step, not a just behavior change.
happy to do it, just flagging the tradeoff first (which doesn't seem necessary, but I've been proven wrong several times here) rather than assuming.

some context for this:
checked against the actual behavior before answering, and `categoryIdOrAlias: string` isn't a mismatch, it's intentional. i.e. this endpoint accepts either an id or an existing alias to identify the category, same as `getDetails`.

"must use id" was specifically about the first assignment (a category with no alias yet has nothing else to identify it by), not a universal rule... retrying the same assignment by addressing the category via its own current alias is also supported and tested (should resolve the target category by its existing alias, not only by numeric id, `categoryRouter.spec.ts`)

packages/data-provider/src/utils/formatUtils.ts
justincorrigible: update: this changed in a later commit after feedback:
purely numeric values are now rejected outright, and an alias can never collide with a category's own numeric id. a value containing a decimal point (e.g. 2.5) is still allowed, since it can never be mistaken for a whole-number id, and can still serve as a data version label.

